PR #14, opened 2015-01-27
Deprecate "mb" (Fixes JB#26064)

thp:
"mb2" is better than "mb" in many ways. Time to migrate.


rburchell:
I suppose that's better than an outright 'rm' in that it hints where to look, but it won't properly be transitional. mb2 requires a -s if there are multiple spec files present, and also requires a subcommand all the time (like mb2 build) -- just off the top of my head. There may be a few other differences I forget.

jlehtoranta:
They both seem to require a subcommand all the time, but "mb2" doesn't have a "clean" command. "mb" always wants the spec file specified as the last argument, but "mb2" requires a "-s" in case of multiple spec files. So, if this wrapper is used in a common case like "mb build path/to/spec", "mb2" will ignore the spec file path and auto-search the spec file. I suppose it will work out in most cases though. Besides that the build results of "mb" end up under "~/rpmbuild", whereas "mb2" will build in the current directory. The least effort way could be e.g. to add a friendly note about these differences after "'mb' is obsolete, please use 'mb2'". I can merge this after that, if there aren't any further opinions on making "mb" obsolete.

rburchell:
I have no objections, that was just a case of musing out loud. I'd be happy to see it happen.

thp:
Let's just merge this and get the transitional issues fixed as we experience them? As soon as a dev complains about "you broke my mb!" we can educate them on how to use mb2.

I will update the patch to not call mb2 with arguments, but just show an error message.

rburchell:
May you rest in pieces.

PR #16, opened 2015-09-05
Fix for users using a different shell than bash

simonvanderveldt:
Uses `awk` to replace the shell value in `/etc/passwd` with `/bin/bash`


sledges:
LGTM

simonvanderveldt:
Moved to mer-tools/sdk-setup/pull/35

PR #20, opened 2016-10-19
Allow to run qmake and make without qtc macros. Contributes to JB#34990

martyone:


jpetrell:
LGTM

Thank you! Both mb2 make clean and distclean worked as expected. :)

PR #21, opened 2016-10-20
Jb36486

martyone:


sdk-setup/src/sdk-assistant
```diff
@@ -164,7 +166,11 @@ fi
 
 if [[ $OPT_CREATE ]]; then
     if [[ ! $OPT_TARGET ]] || [[ ! $OPT_ROOTSTRAP ]]; then
-        fail "create: target name and rootstrap required"
+        fail "create: target name and rootstrap URL or filename required"
+    fi
+
+    if [[ ! -f $OPT_ROOTSTRAP ]]; then
+        fail "$OPT_ROOTSTRAP: no such file"
```
saukko: Would be good to define the filename in error message.

martyone: You mean to say "no such rootstrap file"? This could be also interpreted as "the file exists but it is not a valid rootstrap file" - for that reason I prefer the current message.

Please confirm you want to update it.

jlehtoranta:
LGTM

PR #22, opened 2016-10-20
[sdk-foreach] Initial commit. Contribute to JB#36492

martyone:


jlehtoranta:
Excellent, but how about excluding hardware devices by default and having them included only by a separate option?

martyone:
Originally I planned to add (more) options like that but soon I become afraid such a feature would hardly find some users - who deals with more than three devices? :) The -i (interactive) option works well here, don't you think?

jlehtoranta:
Ok, I think the default mode should be the interactive mode then - just to be safe that only users, who intend to run the same command on *all* devices, do it.

jlehtoranta:
lgtm now 👍

PR #23, opened 2016-10-21
Add prep command.

jusa:


sdk-setup/src/mb2
```diff
@@ -31,9 +31,9 @@ short_usage() {
     cat <<EOF
 usage: $ME [-t <target>] [-s <specfile>] [-d <device>] [-p <projectdir>]
            [-f <folder>] [-i] [-P] [-x]
-           build [-d] [-j <n>] [<args>] | qmake [<args>] | make [<args>] | ssh <args>
+           build [-e] [-d] [-j <n>] [<args>] | qmake [<args>] | make [<args>] | ssh <args>
```
rainemak: Why ```-e``` not ```-p``` (lower case).  Could we use same flags as there are in rpmbuild?

See:
http://www.rpm.org/max-rpm-snapshot/ch-rpm-b-command.html

jusa: Because even though the flags do not overlap I didn't want to have -p for both global option (for projectdir) and dospec.. The long argument is still the same as with rpmbuild.

rainemak:
:+1: LGTM

sdk-setup/src/mb2
```diff
@@ -58,9 +58,12 @@ usage() {
   $ME is aware of spectacle and will update the spec file if there is
   an obvious yaml file which is newer.
 
-  $ME build [-d] [-j <n>] [<args>]
+  $ME prep           : runs rpmbuild for %prep phase of the spec file.
+
+  $ME build [-bp|--doprep] [-d|--enable-debug] [-j <n>] [<args>]
                      : runs rpmbuild for the given spec file in the
                        given sb2 target. Produces an rpm package.
+                     : -bp    run %prep phase when building as well
```
martyone: I prefer that single-dash options are also single-letter. Just '-p' would be OK with me.

jlehtoranta:
LGTM

PR #27, opened 2016-11-29
mer-sdk-chroot: Do not pass --preserve-environment to 'su'

martyone:
*Attention:* Temporarily on top of branch 'jb36783', the destination will be changed to 'master' once jb36783 is merged.

This is not only mutually exclusive with --login; its effect also
differs from what the original author apparently thought to be.


jlehtoranta:
lgtm

PR #28, opened 2016-12-01
Jb37077

martyone:


jlehtoranta:
lgtm

martyone:
I will spend some more time on this - it does not work well with non-trivial merging strategies.

PR #40, opened 2017-05-26
[sdk-setup] Increase nproc limit inside SDK chroot

martyone:


martyone:
The default config provided by 'pam' package sets this to 1024.

PR #48, opened 2017-07-13
Jb38303

martyone:


rpm/sdk-setup.spec
```diff
@@ -43,6 +43,16 @@ Conflicts:  sdk-chroot
 %description -n sdk-vm
 Contains the supporting configs for VMs
 
+%package -n sdk-tooling-chroot
+Summary:    Mer SDK files for build tools distribution
+Group:      System/Base
+BuildArch:  noarch
+Requires(pre): rpm
```
pvuorela: Um, can you install rpm without rpm? :)

martyone: I do not know. The author of the original code for sdk-tooling-chroot above might known more, so I prefer keeping it there.

PR #50, opened 2017-08-07
Fix user shell

martyone:


marmistrz:
And `sdk-setup/src/mer-sdk-lxc`? It used getent passwd too.

martyone:
Updated also mer-sdk-lxc but let me note that this hasn't been actively maintained for some time and I am not sure if it even works nowadays.

marmistrz:
I've never used it either, but better safe than sorry :)

PR #67, opened 2018-04-10
Do not produce short-circuited RPMs

martyone:


martyone:
Requires https://git.merproject.org/mer-core/rpm/merge_requests/9

mkosola:
What is short-circuited RPM?

martyone:
@mkosola produced with rpmbuild --short-circuit, tainted with an unsatisfiable dependency on rpmlib(ShortCircuited) - check e.g. https://bugzilla.redhat.com/show_bug.cgi?id=671210

mkosola:
@martyone Ah, thank you!

saukko:
We should add dependency to the rpm version that contains this new option so that one can't by accident use newer mb2 with old rpm.

In general LGTM :+1:

martyone:
@saukko that's a bit tricky as our rpm patching is not reflected in package version, only the OBS-generated release number changes. Checking other packages' specs, none seems to be referring that.

PR #70, opened 2018-05-11
[mb2] Do not fail on modified git submodules. Contribute to JB#41867

martyone:
Also reflect the state of the submodules in the version string.


krnlyng:
works and lgtm :)

PR #73, opened 2018-05-22
Jb41932

martyone:


sdk-setup/src/mb2
```diff
@@ -125,7 +125,9 @@ usage() {
   -P | --pedantic      : do extra checks
   -x[=<tag>] | --fix-version[=<tag>]
                        : this option is implied when used inside a Git working
-                         tree from within a platform SDK (use -X to override).
+                         tree (use -X to override). Within Application SDK it is
+                         only implied if the version recorded in the spec file
+                         is exactly "0".
```
saukko: Why do we say here Within Application SDK? Isn't it implied always when the Version: is set to 0?

saukko: Or more actually shouldn't we imply it always when the Version: is set to 0 that is?

martyone: It does not say "It is only implied in App SDK" - it says "In App SDK it is only implied if ...". OK?

PR #78, opened 2018-06-12
Jb42150

martyone:


martyone:
Note that all except the very last commit belong to PR #77

abranson:
All looks good to me. Should help people avoid a few pitfalls.

PR #79, opened 2018-06-15
 [sdk-make-qmltypes.sh] Initial commit. Contribute to jb#41605 

martyone:


martyone:
Note that the first commit belongs to #75

PR #84, opened 2018-08-21
mb2: Add option to not update target dependencies.

jusa:
Adds option -n or --no-deps

Sometimes it is useful to do the mb2 operations on a target without
updating package dependencies (for example, having installed
locally built version of dependency for testing). Add option to
disable the dependency updating so that target stays in proper state
for the duration of build etc.

For example to build without updating dependencies
mb2 -t target -n build

[mb2] Add option to not update target dependencies. Fixes JB#42728

### DIFF
--- a/sdk-setup/src/mb2
+++ b/sdk-setup/src/mb2
@@ -72,6 +72,7 @@ usage() {
   the actual changelog.
 
   $ME prep           : runs rpmbuild for %prep phase of the spec file.
+                     : can use -n
 
   $ME build [-p|--doprep] [-d|--enable-debug] [-j <n>] [<args>]
                      : runs rpmbuild for the given spec file in the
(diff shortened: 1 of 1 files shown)


martyone:
It is a very common situation, that you have some locally built packages installed, and I do not remember I had ever problems with mb2 replacing my locally installed packages. It happens for you - which way did you build and install those testing packages? (Thinking this PR might not be addressing the root cause.)

jusa:
Built normally with mb2 build, installed with sb2 -t target -m sdk-install -R zypper in package (or rpm -U --force package).
I started to verify this issue but to my surprise couldn't at first reproduce. But, when the package built has either higher version required than the locally installed package, or the package built has absolute version required and locally installed package differs from this, then zypper pulls the repository dependancy (makes sense right). Where this is really issue is with local kernel building/packaging. My assumption is that some part of the pieces needed when packaging kernel have absolute version requirement and they then pull the deps. I didn't delve into the kernel packaging for now, but either there is something in the process that could be changed or then just apply this PR for the rare case of when the versions could mismatch as explained.

sdk-setup/src/mb2
```diff
@@ -257,6 +259,8 @@ ensure_spec_newer_than_yaml() {
 }
 
 verify_target_dependencies() {
+    [[ $OPT_NO_DEPS -eq 1 ]] && return
```
martyone: Just "[[ $OPT_NO_DEPS ]]" without "-eq 1" please.

martyone:
OK, then except the one nitpick LGTM :)

PR #95, opened 2018-11-26
Jb41700

martyone:


sdk-setup/src/mb2
```diff
@@ -557,8 +557,8 @@ run_apply() (
             echo "Reversing $patch_file"
```
vigejolla: Why are reverse patches treated differently, i.e. output goes to stdout

martyone: Short answer didn't work for me, so...

Thinking generally, a program may produce output that is either 1) primary outcome of its execution (e.g. what `ls` normally prints), or 2) describes progress/log of normal operation (e.g. `make` printing the command it executes), or 3) informs of expected, recoverable failure (e.g. when download fails from primary source but a mirror is available) or 4) informs of unexpected error in program execution (e.g. invariant/constraint violation)

Considering the traditional output handling, sending all messages either to stdout or stderr, we are mainly interested in the ability to

A) pipe primary output (case 1) to another program for further processing, or
B) redirect normal log (case 2 and possibly 3) to /dev/null for silent operation, and
C) still be informed on errors (case 4)

For that, normally, if case 1 applies, just those messages should go to stdout; if case 1 does not apply, messages from case 2 and possibly 3 should go to stdout; and messages from case 4 should always go to stderr.

PR #97, opened 2018-12-11
[sdk-utils] add script to execute validators in target

vigejolla:
rpmvalidation.sh is now part of sdk-utils
rpmvaliadtion.sh will execute all validators which are found on target


martyone:
We would also need to

1) override the default set of suites with optional switch `--suites suite1[,suite2]...`

2) list available suites with `--list-suites`. The output would be limited to certain target (also omitting the first column) when the `--target` switch is passed as well:

without `--target`:

  <target> <name> {essential|optional} <description>

with `--target`:

  <name> {essential|optional} <description>

vigejolla:
Selecting the suite is a separate task, so I will create a separate PR for that

sdk-setup/src/rpmvalidation.sh
```diff
@@ -0,0 +1,75 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+usage() {
+    cat <<EOF
+Execute RPM validators for the given RPM file
```
martyone: I would reuse the original description "Run a basic quality criteria check for the given RPM file."

sdk-setup/src/rpmvalidation.sh
```diff
@@ -0,0 +1,75 @@
...
+
+Usage: 
+    $0 -t <target> <rpm-file>
+	
+Options:
+    -t|--target <target> Execute the validators on target.
```
martyone: I would say: "The OS version to validate for. Accepts the same argument as 'sb2 -t' does."

sdk-setup/src/rpmvalidation.sh
```diff
@@ -0,0 +1,75 @@
...
+EOF
+}
+
+run_validators() {
+    VALIDATORS=$(sb2 -t $OPT_TARGET ls /usr/libexec/validators)
```
martyone: It should rather be `/usr/libexec/rpmvalidation/suites`, $OPT_TARGET should be quoted

sdk-setup/src/rpmvalidation.sh
```diff
@@ -0,0 +1,75 @@
...
+    echo "Using validators: $VALIDATORS"
```
martyone: I would say "Using validation suites:"

sdk-setup/src/rpmvalidation.sh
```diff
@@ -0,0 +1,75 @@
...
+	for f in $VALIDATORS; do
+		sb2 -t $OPT_TARGET "/usr/libexec/validators/$f" $@ || break
+	done
+}
+
+while [[ $1 ]]; do
+    case "$1" in
```
martyone: No need to quote. Basic rule: quote every argument to a "command" (unless word splitting is wanted), do not quote any argument to shell language constructs. Check e.g. sdk-motd for reference.

sdk-setup/src/rpmvalidation.sh
```diff
@@ -0,0 +1,78 @@
...
+if [[ $# -lt 3 ]]; then
+    usage
+    exit 1
+fi
+
+VALIDATOR_PARAMS=()
+
+while [[ $1 ]]; do
+    case "$1" in
+        "-t" | "--target")
```
martyone: ditto, also trailing space

sdk-setup/src/rpmvalidation.sh
```diff
@@ -0,0 +1,78 @@
...
+    exit 1
+fi
+
+VALIDATOR_PARAMS=()
+
+while [[ $1 ]]; do
+    case "$1" in
+        "-t" | "--target") 
+            shift
+            OPT_TARGET="$1"
```
martyone: ditto

sdk-setup/src/rpmvalidation.sh
```diff
@@ -0,0 +1,78 @@
...
+            ;;
+        *)
+            VALIDATOR_PARAMS+="$1 "
```
martyone: Use arrays `arr+=("$1")`, then expand with `"${arr[@]}"`

sdk-setup/src/rpmvalidation.sh
```diff
@@ -0,0 +1,78 @@
...
+            ;;
+    esac
+    shift
+done
+
+if [[ "$OPT_TARGET" ]]; then
```
martyone: ditto

sdk-setup/src/rpmvalidation.sh
```diff
@@ -0,0 +1,78 @@
...
+	if sb2 -t $OPT_TARGET [ -d /usr/libexec/validators ]; then
```
martyone: ditto

sdk-setup/src/rpmvalidation.sh
```diff
@@ -0,0 +1,78 @@
...
+        echo "ERROR [$OPT_TARGET] Target does not contain validators. Use a more recent target."
+        echo "FAILED"
+    fi
+else
+    echo "Target"
+    echo "======"
+    echo "ERROR [] Target not specified"
+    echo "FAILED"
+fi
+
```
martyone: trailing blank line

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,104 @@
...
+
+populate_validators() {
+    VALIDATOR_INIS=$(sb2 -t "$OPT_TARGET" ls "$SUITE_INI_DIR")
+    for f in $VALIDATOR_INIS; do
+        validators+=(${f%.ini})
+    done
+}
+
+validator_value() {
+    sb2 -t "$OPT_TARGET" python3 -c 'import sys, configparser; config = configparser.RawConfigParser(); config.read(sys.argv[2:]); sec, val = sys.argv[1].split("."); print(config[sec][val]);' RpmValidationSuite.$2 /usr/share/rpmvalidation/$1.ini
```
martyone: It does not allow overriding the default value yet and the python line is really long :)

Maybe having global variable `CONFIG_DIRS=({/usr/share,/etc}/rpmvalidation/suites)`, and using here something like..?

```
local suite=$1
local key=$2

local ini_files=("${CONFIG_DIRS[@]/%//$suite.ini}")

local python=
read -d '' -r python <<'END'
...
END

sb2 -t "$OPT_TARGET" python3 -c "$python" "$key" "${ini_files[@]}"
```

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,104 @@
...
+}
+
+run_validator() {
+    validator_exec=$(validator_value $1 exec)
+    validator_cmd=$(echo "$validator_exec" | sed -e "s#%f#$RPM_FILE#" -)
```
martyone: `suite_exec=${suite_exec//%f/$RPM_FILE}`

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,104 @@
...
+        validators+=(${f%.ini})
+    done
+}
+
+validator_value() {
+    sb2 -t "$OPT_TARGET" python3 -c 'import sys, configparser; config = configparser.RawConfigParser(); config.read(sys.argv[2:]); sec, val = sys.argv[1].split("."); print(config[sec][val]);' RpmValidationSuite.$2 /usr/share/rpmvalidation/$1.ini
+}
+
+run_validator() {
+    validator_exec=$(validator_value $1 exec)
```
martyone: Please prefer always naming function arguments. BTW I would prefer using CamelCase for the keys so that we do similar as e.g. XDG and systemd. I will also use the word "suite" in the snippet as I suggested before:

```
local suite=$1
suite_exec=$(suite_value "$suite" Exec)
```

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,104 @@
...
+    validator_cmd=$(echo "$validator_exec" | sed -e "s#%f#$RPM_FILE#" -)
+    sb2 -t "$OPT_TARGET" "$validator_cmd"
+}
+
+run_all_validators() {
+    echo "Using validation suites: ${validators[@]}"
+	for f in ${validators[@]}; do
```
martyone: The expansion should be quoted `"${var[@]}"`, otherwise the "@" has no effect compared to "*". Also `f` should be declared as local beforehand. And it is a good convention to always initialize variable when declaring it as local - in some more complex cases the behavior is surprising so I learned to always initialize it.

```
local f=
for f in ...; do
```

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,104 @@
...
+if [[ $0 == *.sh ]]; then
+    echo "Warning: The name '$0' is deprecated. Use '${0%.sh}' instead"
+fi
+
+if [[ $# -lt 3 ]]; then
+    usage
+    exit 1
+fi
+
+
```
martyone: Extra blank

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,104 @@
...
+run_all_validators() {
+    echo "Using validation suites: ${validators[@]}"
+	for f in ${validators[@]}; do
+        run_validator $f
+	done
+}
+
+if [[ $0 == *.sh ]]; then
+    echo "Warning: The name '$0' is deprecated. Use '${0%.sh}' instead"
```
martyone: This should go to stderr

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,104 @@
...
+        *)
+            RPM_FILE=$1
+            ;;
+    esac
+    shift
+done
+
+populate_validators
+
+if [[ $OPT_TARGET ]]; then
```
martyone: I would handle the error conditions first and implement the normal execution path at the very end of the file. The error messages need a bit more love :)

The script may grow in future - it is a good convention to encapsulate the "main" code to three functions `set_defaults`, `parse_options` and `main`, which allows easier debugging and test writing later (you can e.g `source` (`.`) the script in an interactive shell and debug each function without invoking the main code). Check e.g. sdk-motd for reference.

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,206 @@
...
+    if [[ ! $OPT_TARGET ]]; then
+        echo "Target"
+        echo "======"
+        echo "ERROR [] Target not specified"
+        echo "FAILED"
+        return
+    fi
+
+    echo "Running validation suites on target $OPT_TARGET"
+    run_all_suites
```
martyone: It should be `run_essential_suites` instead. And we should also add the `--suites suite1[,suite2]...` switch to allow overriding this.

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,206 @@
...
+SUITES=()
+
+populate_suites() {
+    local target=$1
+    local validator_inis=()
+    SUITES=()
+    if [[ ! $target ]]; then
+        return
+    fi
+    validator_inis=($(sb2 -t "$target" ls "$DEFAULT_INI_DIR" 2>/dev/null)) || return
```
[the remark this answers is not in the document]
vigejolla: I don't think there should be an ini file in /etc without corresponding ini file in /usr/share

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,206 @@
...
+    sb2 -t "$OPT_TARGET" python3 -c "$python" "RpmValidationSuite.$key" "${ini_files[@]}"
+}
+ 
+list_target() {
+    local target=$1
+    local display_target=$2
+    local f=
+    for f in "${SUITES[@]}"; do
+        local essential=$(suite_boolean "$target" "$f" Essential)
+        if [[ "$essential" == "True" ]]; then
```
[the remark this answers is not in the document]
vigejolla: Is there any harm in quoting? To me it makes it clearer that we are dealing with strings here.

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,206 @@
...
+}
+
+parse_options() {
+    if [[ $0 == *.sh ]]; then
+        echo "Warning: The name '$0' is deprecated. Use '${0%.sh}' instead" >&2
+    fi
+
+    while [[ $1 ]]; do
+        case $1 in
+            "-h" | "--help")
```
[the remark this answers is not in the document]
vigejolla: Again I don't think it's unnecessary because it helps the reader understand that we are dealing with strings here.

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,206 @@
...
+        usage
+        return
+    fi
+
+    populate_suites $OPT_TARGET
+
+    if [[ $OPT_TARGET ]]; then
+        if [[ ${#SUITES[@]} -lt 1 ]]; then
+            echo "Target"
+            echo "======"
```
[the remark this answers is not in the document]
vigejolla: No. QtC uses it in the Issues pane.

sdk-setup/src/rpmvalidation
vigejolla: The task specifically was to execute ALL suites.

sdk-setup/src/rpmvalidation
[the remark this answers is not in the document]
martyone: That's the common way of use. /usr/share is for distribution packages, /etc is for host specific configuration - which may override features provided by distribution packages as well as add new features.

sdk-setup/src/rpmvalidation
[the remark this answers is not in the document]
martyone: It's unnecessary. In many other cases quotes are inevitable and not many people are fully aware of the implications - using them (only) where they are needed is a clear message that you understand these implications :)

sdk-setup/src/rpmvalidation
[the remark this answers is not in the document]
martyone: when you use shell interactively, do you use it like in

`ls "-l" "Documents/"`

or like in?

`ls -l Documents/`

It's the same language. There are cases where quotes are inevitable and other cases where they are optional - let's make a clear impression we know what we do and save us from some extra typing :)

sdk-setup/src/rpmvalidation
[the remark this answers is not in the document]
martyone: What do you mean more exactly? Shoudn't we comment here then?

sdk-setup/src/rpmvalidation
martyone: I wouldn't spent too much time on getting an ideal task granularity and try to finish it finally - it's a trivial difference on about 5 lines of code if I omit help text :)

sdk-setup/src/rpmvalidation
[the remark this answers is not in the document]
vigejolla: I don't care what people think of my understanding, to me it's more important that they understand the code. In my opinion it is more readable with quotes.

sdk-setup/src/rpmvalidation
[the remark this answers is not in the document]
vigejolla: Well, I do write ls -l "My Documents/". But that's really not the point. Here "-h" and "--help" are strings that I'm parsing. They are not part of the case syntax, like | and ).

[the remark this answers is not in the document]
vigejolla: .. and you have definitely typed more characters in these comments than my extra quotes :)

sdk-setup/src/rpmvalidation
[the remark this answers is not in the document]
vigejolla: I'll add a comment

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,229 @@
...
+    local suite=
+    for suite in "${SUITES[@]}"; do
+        run_suite $suite
+    done
+}
+
+run_selected_suites() {
+    OLD_IFS=$IFS
+    IFS=','
+    SUITES=($OPT_SUITES)
```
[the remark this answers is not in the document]
vigejolla: Modifying OPT_SUITES after parsing the options seems like a nasty thing to do. But let's see if I can improve this.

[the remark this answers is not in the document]
vigejolla: Now OPT_SUITES is parsed using read -a like you suggested. I also modified populate_suites so that it checks that the given suites actually exist on target.

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,240 @@
...
+    local target=$1
+    local validator_inis=()
+    local uniquie_inis=()
+    SUITES=()
+    if [[ ! $target ]]; then
+        return
+    fi
+    for d in ${CONFIG_DIRS[@]}; do
+        validator_inis+=($(sb2 -t "$target" ls "$d" 2>/dev/null))
+    done
```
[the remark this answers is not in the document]
vigejolla: that doesn't work because ls prints the directory names as well. But I modified the idea slightly, and use find instead of ls.

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,240 @@
...
+}
+
+suite_raw() {
+    suite_value $1 $2 $3 raw
+}
+ 
+list_target() {
+    local target=$1
+    local display_target=$2
+    local f=
```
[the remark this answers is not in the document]
vigejolla: :(

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,235 @@
...
+    done
+    return 0
+}
+
+populate_suites() {
+    local target=$1
+    local validator_inis=()
+    local uniquie_inis=()
+    SUITES=()
+    validator_inis=($(sb2 -t "$target" find "${CONFIG_DIRS[@]}" -type f -printf %f\\n 2>/dev/null))
```
martyone: Yup, good idea. It would break on ini file under subdir though, so we should either add `-maxdepth 1` or use `%P` instead of `%f`. We could also add `-name "*.ini"`. Feel free to use quotes instead of escaping :)

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,235 @@
...
+    suite_exec=$(suite_raw "$OPT_TARGET" "$suite" Exec)
+    suite_cmd=${suite_exec//%f/$OPT_RPM_FILE}
+
+    sb2 -t "$OPT_TARGET" "$suite_cmd"
+}
+
+run_suites() {
+    echo "Using validation suites: ${SUITES[@]}"
+    local suite=
+    for suite in ${SUITES[@]}; do
```
martyone: quotes :) similar below and one or two occurrences elsewhere

sdk-setup/src/rpmvalidation
```diff
@@ -0,0 +1,238 @@
...
+    parse_options "$@" || return
+
+    if [[ $OPT_USAGE ]]; then
+        usage
+        return
+    fi
+
+    if [[ $OPT_TARGET ]]; then
+        populate_suites "$OPT_TARGET"
+        if [[ "${#SUITES[@]}" -lt 1 ]]; then
```
martyone: We should not fail here for the `--list-suites` command, so

`if [[ "${#SUITES[@]}" -lt 1 && ! $OPT_LIST ]]; then`

PR #99, opened 2018-12-21
Jb39150

jlehtoranta:
Make connman services mutable for allowing to change the proxy configuration

### DIFF
--- a/sdk-setup/src/sdk-setup-enginelan
+++ b/sdk-setup/src/sdk-setup-enginelan
@@ -7,7 +7,7 @@
 DEVICES_XML=/etc/mersdk/share/devices.xml
 
 DEF_SUBNET="10.220.220"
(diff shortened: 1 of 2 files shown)


sdk-setup/src/sdk-setup-enginelan
```diff
@@ -58,24 +58,17 @@ else
...
-Description = SDK Engine Configuration
 
-[service_sailfishsdk_ethernet]
-Type = ethernet
-# TODO: The desired effect here is to prevent connman setting gateway - how to do that properly?
-IPv4 = $subnet.254/24/$subnet.253
-IPv6 = off
-MAC = 08:00:5A:11:F1:55
+con_sailfishsdk="ethernet_08005a11f155_cable"
+con_nat="ethernet_"$MAC2"_cable"
```
martyone: `${MAC2}` would do the trick :)

jlehtoranta: Fixed, thanks!

PR #102, opened 2019-01-30
[mb2] Use sshport if it is defined in devices.xml. Contributes to JB#43485

vigejolla:


sdk-setup/src/mb2
```diff
@@ -434,21 +434,24 @@ get_device() {
     if [[ "$device_subnet" ]]; then
         device_ip="$device_subnet".$device_index
     fi
+    if [[ -z "$device_sshport" ]]; then
```
martyone: Please prefer using `!` instead of `-z` - compare with the test for device_subnet above

vigejolla: Why?

vigejolla: I already changed it, but I would still like to know why is ! preferred over -z

martyone: I meant check the line a bit above, that it similarly uses `[[ "$device_subnet" ]]`, instead of `[[ -n "$device_subnet" ]]`. There are just few cases that are inconsistent on this. (And yes, the quotes are unnecessary, but there is total disorder regarding quotes in this source, so.. :))

PR #103, opened 2019-02-04
Jb41701

martyone:


sdk-setup/src/sdk-manage
```diff
@@ -1277,6 +1365,155 @@ install_target() {
     succeeded=1
 }
 
+snapshot_target() {
+    local reset= positional_args=()
+    while [[ $# -gt 0 ]]; do
+        case $1 in
+            -r|--reset)
```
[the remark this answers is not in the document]
martyone: It provides an EAFP style UI for two-step operation: first verify that the original target matches the one from which the snapshot was taken initially, then reset. That's why the original target is required. This command is not really meant to be used by human user under normal conditions, and if we considered to make the original target optional when '--reset' is used (skipping the verification step), the help text would have to change from the concise

```
snapshot [--reset] <original> <snapshot>
```

to

```
snapshot <original> <snapshot>
snapshot --reset [<original>] <snapshot>

```

and similarly for the long description, so I would go with both arguments mandatory :)

martyone:
Added --no-refresh to installdeps, implied by --snapshot. Added optional verbose|quiet argument to --search-output-dir, so that one can assess what was installed from the output dir. Ensured that packages that are already installed in the target are upgraded from output dir.

sdk-setup/src/mb2
```diff
@@ -157,8 +157,17 @@ GLOBAL OPTIONS
         Specify the device
 
     -o, --output-dir <dir>
-        Where to place produced RPM files. Defaults to ./RPMS directory.  With
-        this option existing files in the output directory are preserved
+        Defaults to './RPMS'. The resulting RPM packages will be placed
```
vigejolla: I wonder if it could be stated a little bit more explicit that the order of options matter. "-o RPMS --no-search-output-dir" is not equal to "--no-search-output-dir -o RPMS".

martyone: Done + a little more help polishing

PR #109, opened 2019-02-27
[rpmvalidation] Fix busybox compatibility under sb2. Fixes JB#44979

martyone:
Busybox's find does not recognize '-printf'.

Also eliminated the suppression of find's stderr - hidding the error
message made debugging this harder.

### DIFF
--- a/sdk-setup/src/rpmvalidation
+++ b/sdk-setup/src/rpmvalidation
@@ -74,7 +74,9 @@ populate_suites() {
     local suite_inis=()
     local unique_inis=()
     SUITES=()
-    suite_inis=($(sb2 -t "$target" find "${CONFIG_DIRS[@]}" -maxdepth 1 -name "*.ini" -type f -printf "%f\n" 2>/dev/null))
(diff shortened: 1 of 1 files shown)


sdk-setup/src/rpmvalidation
```diff
@@ -74,7 +74,9 @@ populate_suites() {
     local suite_inis=()
     local unique_inis=()
     SUITES=()
-    suite_inis=($(sb2 -t "$target" find "${CONFIG_DIRS[@]}" -maxdepth 1 -name "*.ini" -type f -printf "%f\n" 2>/dev/null))
+    local query=$(printf 'cd %q 2>/dev/null \
```
vigejolla: Why is cd needed here? Wouldn't just simple 'find %q -maxdepth 1 -name "*.ini" -type f -print\n' work just as well?

martyone: > Also eliminated the suppression of find's stderr - hidding the error message made debugging this harder.

:)

PR #110, opened 2019-03-04
Jb43687

martyone:


sdk-setup/src/resize-rootfs
```diff
@@ -1,47 +1,58 @@
...
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+set -o errexit
+set -o pipefail
```
vigejolla: I believe pipefail is a bash extension

martyone: Indeed :(

sdk-setup/src/resize-rootfs
```diff
@@ -1,47 +1,65 @@
...
-   rm /resize_root_fs	
-   exit
-fi
+set -o errexit
 
-pre=$(mktemp)
-post=$(mktemp)
+pipefail()
+{
+    case "${PIPESTATUS[*]}" in
```
vigejolla: Unfortunately also PIPESTATUS is bash only :(
There are a few ugly solutions, e.g. here: https://stackoverflow.com/questions/8833396/pipe-command-output-but-keep-the-error-code

martyone: That drives me crazy... :) thanks again

PR #114, opened 2019-03-15
sdk-manage: Compare snapshot to recent uninstalled

ix5:
In addition to recently **installed/updated** packages, also check the build target for recently **uninstalled** packages when comparing with the known snapshot timestamp.

Caveat:
Parsing `/var/log/zypp/history` is ugly and might break in the future.

### DIFF
--- a/sdk-setup/src/mb2
+++ b/sdk-setup/src/mb2
@@ -333,8 +333,9 @@ GLOBAL OPTIONS
 KNOWN ISSUES
 
 Build target snapshot is not reset automatically when the original target is
-changed just by removing packages. Package installation or update is needed to
-trigger the automatic reset.
+changed just by removing packages via plain RPM usage. Either use zypper for
+all package operations or use 'mb2 build-requires reset' to ensure snapshot is
+reset.
 EOF
 }
 
--- a/sdk-setup/src/sdk-manage
+++ b/sdk-setup/src/sdk-manage
@@ -1578,23 +1578,39 @@ target_snapshot_needs_reset() {
 
     local snapshot_time=
     snapshot_time=$(object_config_get "target:$snapshot" "$SNAPSHOT_TIME_KEY") || return
+    # Convert date to unix seconds
+    snapshot_time=$(date --date "$snapshot_time" +%s) || return
 
     # Try to avoid querying RPM database as it takes considerable amount of time
     tmpfile=$(mktemp) || return
-    touch --date "$snapshot_time" "$tmpfile" || return
+    # --date "@value" passes value as unix seconds
+    touch --date "@$snapshot_time" "$tmpfile" || return
     local newer=
     newer=$(find "$MER_TARGETS/$original/var/lib/rpm" -newer "$tmpfile" -print -quit) || return
     if [[ ! $newer ]]; then
         return 1
     fi
 
-    # TODO how to determine recent package removal?
     local most_recent_install_time=
     most_recent_install_time=$(set -o pipefail; rpm --root="$MER_TARGETS/$original" -qa \
         --queryformat '%{installtime}\n' |sort -n |tail -n1) || return
-    local snapshot_time_s=$(date --date "$snapshot_time" +%s) || return
+    # Compare values in unix seconds
+    (( most_recent_install_time > snapshot_time )) && return
 
-    (( most_recent_install_time > snapshot_time_s ))
+    # The rpm database does not know about the most recent package removal, so
+    # parse the zypper install log. zypp/history looks like this:
+    # 2019-02-27 16:39:18|remove |droid-hal-mydevice-tools|0.0.6-201902271629|armv7hl|user@pc|
+    # But could also look like this:
+    # 2019-02-27 16:39:18|install|remove-my-pkg|0.0.6-201902271629|armv7hl|user@pc|
+    local most_recent_removal_time=$(set -o pipefail; \
+        tac "$MER_TARGETS/$original/var/log/zypp/history" \
+        |cut -d "|" -f "1,2" \
+        |grep -m 1 "|remove" \
+        |cut -d "|" -f 1 \
+    ) || return
+    most_recent_removal_time=$(date --date "$most_recent_removal_time" +%s) || return
+    # Compare values in unix seconds
+    (( most_recent_removal_time > snapshot_time )) && return
 }
 
 reserve_target() {


sdk-setup/src/sdk-manage
```diff
@@ -1565,13 +1565,23 @@ target_snapshot_needs_reset() {
         return 1
     fi
 
-    # TODO how to determine recent package removal?
     local most_recent_install_time=
     most_recent_install_time=$(set -o pipefail; rpm --root="$MER_TARGETS/$original" -qa \
-        --queryformat '%{installtime}\n' |sort -n |tail -n1) || return
-    local snapshot_time_s=$(date --date "$snapshot_time" +%s) || return
+        --queryformat '%{installtime}\n' | sort -n | tail -n1) || return
```
martyone: That |is |the |style |used |here. Please keep it and follow in the new code :)

sdk-setup/src/sdk-manage
```diff
@@ -1565,13 +1565,23 @@ target_snapshot_needs_reset() {
...
+
+    # The rpm database does not know about the most recent package removal, so
+    # parse the zypper install log. zypp/history looks like this:
+    # 2019-02-27 16:39:18|remove |droid-hal-mydevice-tools|0.0.6-201902271629|armv7hl|user@pc|
+    local most_recent_uninstall_time=$(set -o pipefail; \
+        tac $MER_TARGETS/$original/var/log/zypp/history \
```
martyone: should be quoted

sdk-setup/src/sdk-manage
```diff
@@ -1565,13 +1565,23 @@ target_snapshot_needs_reset() {
...
+        |  grep -m 1 "|remove" \
+        | cut -d "|" -f 1 \
+    ) || return
+    most_recent_uninstall_time=$(date --date "$most_recent_uninstall_time" +%s) || return
 
-    (( most_recent_install_time > snapshot_time_s ))
+    # Compare values in unix seconds
+    local snapshot_time_s=$(date --date "$snapshot_time" +%s) || return
+    (( most_recent_install_time > snapshot_time_s ) || ( most_recent_uninstall_time > snapshot_time ))
```
martyone: most_recent_uninstall_time > snapshot_time**_s**.

Also I do not believe this syntax would be correctly parsed by bash - ever tested it? :)

Either use ``(( i > s || r > s ))`` (notice quotes usage) or better stick snapshot_time_s definition to snapshot_time definition and then test each separately like in `(( most_receint_install_time > snapshot_time_s )) && return`

sdk-setup/src/sdk-manage
```diff
@@ -1565,13 +1565,23 @@ target_snapshot_needs_reset() {
...
-    local snapshot_time_s=$(date --date "$snapshot_time" +%s) || return
+        --queryformat '%{installtime}\n' | sort -n | tail -n1) || return
+
+    # The rpm database does not know about the most recent package removal, so
+    # parse the zypper install log. zypp/history looks like this:
+    # 2019-02-27 16:39:18|remove |droid-hal-mydevice-tools|0.0.6-201902271629|armv7hl|user@pc|
+    local most_recent_uninstall_time=$(set -o pipefail; \
+        tac $MER_TARGETS/$original/var/log/zypp/history \
+        |  grep -m 1 "|remove" \
+        | cut -d "|" -f 1 \
```
martyone: This would also match `2019-02-27 16:39:18|install |remove-foo|1.2.3|armv7hl|user@pc|`

Consider something line `sed -n 's/^\([^|]\+\)|remove.*/\1/p'` instead of grep+cut here.

sdk-setup/src/mb2
```diff
@@ -298,13 +298,6 @@ GLOBAL OPTIONS
         regular use - create a script file named after the .spec file but with
         '.changes.run' extension to instruct $ME to generate changelog with
         'git-change-log <args>' instead.
-
-
-KNOWN ISSUES
-
-Build target snapshot is not reset automatically when the original target is
-changed just by removing packages. Package installation or update is needed to
-trigger the automatic reset.
```
martyone: The solution does not cover plain rpm usage, so the note should be only updated, not fully removed.

ix5: Updated note, now reads:
> Build target snapshot is not reset automatically when the original target is
changed just by removing packages via rpm(only zypper keeps a log of removals).
> Either use zypper for all package operations or a package installation or
update is needed to trigger the automatic reset.

sdk-setup/src/sdk-manage
ix5: Used this format instead(note the additional `cut`):
```
tac ... \
|cut -d "|" -f "1,2" \
|grep -m 1 "|remove" \
|cut -d "|" -f 1 \
```
I've timed 100 runs against your `sed` method and it's a tad faster, 1.3s(non-sed) vs 1.6s(sed).

sdk-setup/src/sdk-manage
ix5: Used `snapshot_time_s` everywhere and also picked up your suggestion so that it now read:
```
(( most_recent_install_time > snapshot_time_s )) && return
(( most_recent_removal_time > snapshot_time_s )) && return
```

sdk-setup/src/sdk-manage
ix5: Done

sdk-setup/src/sdk-manage
ix5: Updated to keep the style

sdk-setup/src/sdk-manage
martyone: ok :)

sdk-setup/src/sdk-manage
```diff
@@ -1565,13 +1565,27 @@ target_snapshot_needs_reset() {
...
+        tac "$MER_TARGETS/$original/var/log/zypp/history" \
+        |cut -d "|" -f "1,2" \
+        |grep -m 1 "|remove" \
+        |cut -d "|" -f 1 \
+    ) || return
+    most_recent_removal_time=$(date --date "$most_recent_removal_time" +%s) || return
+
+    # Compare values in unix seconds
+    local snapshot_time_s=$(date --date "$snapshot_time" +%s) || return
+    (( most_recent_install_time > snapshot_time_s )) && return
```
martyone: We could also test each value immediately after it is determined - if it fails, no need to determine the next one. I would consider eliminating `snapshot_time_s` by storing the seconds directly in `snapshot_time` (prepend `@` when passing to `touch --date`).

sdk-setup/src/mb2
martyone: I would suggest: "[...] just by removing packages via plain RPM usage. Either use zypper for all package operations or use 'mb2 installdeps --reset' to ensure a snapshot is reset.

sdk-setup/src/sdk-manage
ix5: Changed to compare and return on success early.

I have also incorporated your change regarding converting to unix seconds earlier, but imo it does not abet readability and does not really offer any speed benefits. (I really had to dig a bit to find out what `@` does in the context of `date` formatting, and it's a non-standard GNU extension.)

sdk-setup/src/mb2
ix5: Done

sdk-setup/src/sdk-manage
martyone: Yeah, but how usable is the *nix world without GNU extensions? :) I really didn't think about speed benefits - I wanted to bring the rest of the code closer to your approach where you reuse the same variable `most_recent_removal_time` for different "data type" :)

sdk-setup/src/mb2
```diff
@@ -303,8 +303,9 @@ GLOBAL OPTIONS
 KNOWN ISSUES
 
 Build target snapshot is not reset automatically when the original target is
-changed just by removing packages. Package installation or update is needed to
-trigger the automatic reset.
+changed just by removing packages via plain RPM usage. Either use zypper for
+all package operations or use 'mb2 installdeps --reset' to ensure snapshot is
```
martyone: The UI has changed recently, please update instructions to use `mb2 build-requires reset` instead.

ix5: Added https://github.com/sailfishos/sdk-setup/pull/114/commits/9ef37f8ff8904f65b0101d85cd9f5d9491f663bc

ix5:
Rebased

martyone:
@ix5 I would prefer squashing the two commits into one. Please use the "[sdk-manage]" prefix for the original message while doing so.

ix5:
Done

martyone:
Just FYI, no need to mention 'Also change comment from "installdeps" to "build-requires"...', because this is not what the commit does in the end. Little issue - merging anyway. Thanks :)

ix5:
Thanks for the code review!

PR #119, opened 2019-04-15
[zypp] Redirect the new zypp cache location in /home back to /var/cache. Fixes JB#45318

abranson:
/home is shared between sdk engine and targets


abranson:
Tested it with existing and new targets and it seems ok.

PR #132, opened 2019-06-28
[mb2] Fix use with %autosetup. JB#45265

martyone:


martyone:
> LGTM. I assume you have tested this well :)

yeah, we will know soon

PR #136, opened 2019-10-15
Jb47487

vigejolla:


rpm/sdk-setup.spec
```diff
@@ -225,6 +226,10 @@ chmod 755 %{buildroot}%{_bindir}/*
 
 %pre -n sdk-chroot
 if ! rpm --quiet -q ca-certificates && [ -d /%{_sysconfdir}/ssl/certs ] ; then echo "Cleaning up copied ssl certs. ca-certificates should now install"; rm -rf /%{_sysconfdir}/ssl/certs ;fi
+rm -Rf /home/.zypp-cache
```
abranson: If we decide later to clean this up where it was created, it might go away. Should this check for existence before removing?

vigejolla:
> I think it might be better to check for the existence of /home/.zypp-cache before removing it, as it might later get cleaned up by whatever creates it. But it's not important right now.

That's why there is the -f flag in the rm command

martyone:
> I would consider filing a separate bug for the zypp-cache fix.

...so that it can be linked to the bug Andrew created today

vigejolla:
> > I would consider filing a separate bug for the zypp-cache fix.
> 
> ...so that it can be linked to the bug Andrew created today

Yeah, I guess that's best. I'll keep the fix in this PR though, I'll just add a mention to the new bug to the commit message

PR #143, opened 2019-11-22
[sdk-setup] Remove 95-nproc-sdk.conf. JB#48179

vigejolla:
This reverts commit 0bdce56469883a9576d42d06e77771548224cdc4.


vigejolla:
> I would try to improve the message a bit. Maybe saying "remove" and explaining on the second line that another package.... would do better job.

Better now?

martyone:
Yup, you should sync PR description with this - it doesn't do so automatically

PR #146, opened 2020-01-09
[sdk-assistant] Dump help for all unknown commands. Fixes JB#48222

vigejolla:


sdk-setup/src/sdk-assistant
```diff
@@ -192,7 +193,7 @@ while [[ ${1:-} ]]; do
             usage quit
             ;;
         -* )
-            fail "unknown option $1"
+            fail "unknown option $1. Use --help to display usage."
```
rainemak: What about just dumping the help after "unknown option $1"?

vigejolla: > What about just dumping the help after "unknown option $1"?

I thought about that as well and tried it first. The help is quite long, so the reason for the failure ("unknown option $1") would not be visible without scrolling up. That's why I chose option b, i.e. hint how to get help.

rainemak: What if help is printed out first, then empty line, and only after that "unknown option $1"

vigejolla: Good idea. I changed also the "unknown command" case to work in a similar fashion.

martyone: As noted in bz it is uncommon to mess up whole screen on bad usage and I hate the few programs that behave so.. Everyone makes a typo sometimes.

I would prefer to follow the well behaving ones:

```
$ cp --foo
cp: unrecognized option '--foo'
Try `cp --help' for more information.
$ zypper --foo
Unknown option '--foo'
$ gcc --foo
bash: gcc: command not found
$ sb2 gcc --foo
i486-meego-linux-gnu-gcc: error: unrecognized command line option ‘--foo’
i486-meego-linux-gnu-gcc: fatal error: no input files
compilation terminated.
$ osc --foo
osc: no such option: --foo
Try 'osc help' for info.
```

vigejolla:
Returned to just telling how to get help

sdk-setup/src/sdk-assistant
rainemak:
```
$ sdk-assistant
FAIL: command expected
```
With this kind of tool that doesn't work without a command I think that dumping help for instance in above case would be perfectly good. I understand your concern and I'd not like either if cp/mv would dump some help always when doing a typo but as sdk-assistant is a utility tool. Hence, I'd rather compare it to dnf, zypper and git kind of programs. Further, sdk-assistant is not daily tool for developer per se. Even updating target is not necessary to do that often for most of the developer => people forget these. Hence, I think dumping help is a good idea for it.

martyone: yup, I can live with it dumping help when no argument is given :)

vigejolla: It will now print the --help hint always with fail

martyone: @vigejolla Raine wished to get the full help dumped when no argument is given. Would you add that?

vigejolla: > @vigejolla Raine wished to get the full help dumped when no argument is given. Would you add that?

sure (I'll do it later today)

vigejolla: done

sdk-setup/src/sdk-assistant
```diff
@@ -36,6 +37,7 @@ set -o nounset
 
 fail() {
     echo "FAIL: $@"
```
rainemak: something like:
```
echo "unknown option: $@"
```
Sounds less fatal but I can live with these as well.

sdk-setup/src/sdk-assistant
```diff
@@ -36,6 +37,7 @@ set -o nounset
...
+    echo "Try 'sdk-assistant --help' for usage information"
```
martyone: This and the above echo should write to stderr

sdk-setup/src/sdk-assistant
```diff
@@ -210,7 +212,7 @@ case ${1:-} in
         ;;
 esac
 
-[[ $# -gt 0 ]] || fail "command expected"
+[[ $# -gt 0 ]] || usage quit
```
martyone: I would prefer to still say "command expected" or so in addition to dumping the help. We should also write to stderr and exit with non-zero here.

sdk-setup/src/sdk-assistant
martyone: If we remove the prefix, then the messages about missing argument to --user and --pasword option should be changed to not start with the option name.

vigejolla: > If we remove the prefix, then the messages about missing argument to --user and --pasword option should be changed to not start with the option name.

Any suggestions for better error messages in these cases? Or should I just prepend "FAIL: " to those?

vigejolla: After trying this out, I came to conclusion that not having the prefix feels a bit weird too. Maybe just "error:" instead of "FAIL:" - it would not have quite as fatal feeling, but still a bit of "something went wrong" feeling is present.

martyone: "sdk-assistant: " would work well for me as the prefix.

rainemak: Same here.

vigejolla: The obvious choice :)

vigejolla: Replaced the prefix with "sdk-assistant: ". I also changed the prefix in the case when no command is given to match.

sdk-setup/src/sdk-assistant
```diff
@@ -210,7 +217,7 @@ case ${1:-} in
         ;;
 esac
 
-[[ $# -gt 0 ]] || fail "command expected"
+[[ $# -gt 0 ]] || usage "Command expected"
```
martyone: In this case the help text is still not written to stderr while it should be. Better would be dropping the "exit with error if any argument is given" functionality from usage() and let the caller side care about:

`[[ $# -gt 0 ]] || { usage >&2; fail "Command expected; }`

vigejolla: Ok, I can do that as well. I thought you specifically wanted only the error messages go to stderr. The help text is not really error message.

vigejolla: except that doing it the way you suggest would be silly as it would print "Try 'sdk-assistant --help' for usage information" right below the usage information...

vigejolla:
Also the usage goes to stderr now, when run without arguments

PR #147, opened 2020-01-21
[mb2] Allow to select packages to deploy. JB#45576

martyone:


martyone:
> LGTM, even though I would omit the "Qt Creator should be..." part of the commit message. Whether Qt Creator behaviour should be changed or not is something that should be discussed somewhere else than in commit messges.

It explains *why* it is done this ugly way. Read it as:

Proper solution would be to fix Qt Creator to invoke the "package" and/or "make-install" steps
explicitly when needed. We do not have time to this it now, so keep it implied here when running behind Qt Creator.

PR #148, opened 2020-01-22
Jb48118

jlehtoranta:


sdk-setup/src/mb2
```diff
@@ -1194,6 +1261,84 @@ run_prepare() {
     return $rc
 }
 
+cmake_enable_host_paths() {
+    # Handle path changes for Qt Creator
+    if [[ -f "CMakeCache.txt"
```
martyone: I would prefer implementing the guard on caller side. Similarly for cmake_disable_host_paths.

sdk-setup/src/mb2
```diff
@@ -1194,6 +1261,84 @@ run_prepare() {
     return $rc
 }
 
+cmake_enable_host_paths() {
```
martyone: Could we use some more generic name for the two functions? Something that could be understood just that some unspecified setup/cleanup is done for cmake.

sdk-setup/src/mb2
```diff
@@ -1001,6 +1008,66 @@ END
...
+            done
+            set -- "${args[@]}"
+        fi
+
+        assignments=()
+        for arg in "$@"; do
+            if [[ $arg =~ ^[a-zA-Z_][a-zA-Z0-9_]*= ]]; then
+                assignments+=("$arg")
+            fi
+        done
```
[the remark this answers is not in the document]
vigejolla: > Unused.
> 
> BTW, is there going to be similar feature for cmake with the variables cache that we have for qmake?

Removed (for now). Let's think about the variables cache later (if needed)

sdk-setup/src/mb2
```diff
@@ -1194,6 +1261,84 @@ run_prepare() {
...
+        for cbp_file in *.cbp.qtccache; do
+            if [[ -f $cbp_file ]]; then
+                mv $cbp_file ${cbp_file%.qtccache}
+            fi
+        done
+    fi
+}
+
+run_cmake() {
+    local rpmbuild="--build . --target all"
```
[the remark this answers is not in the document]
vigejolla: I agree the name is not very accurate/descriptive. But then again I couldn't figure out what would be better. parameters_which_specify_that_we_should_run_rpmbuild is a bit long...

sdk-setup/src/mb2
```diff
@@ -1634,10 +1779,26 @@ if [[ $1 != --self-test ]]; then  ###  M A I N  EXECUTION BEGINS HERE  #########
...
 handle_pro_file_or_dir_option() {
     local pro_file_or_dir=$1
+    local cmakecache_file="CMakeCache.txt"
+    local read_from_cmakecache=0
 
     rm -f "$STATEDIR/src"
 
-    if [[ ! $pro_file_or_dir ]]; then
+    # Read the project source path from the CMakeCache file, if the path hasn't
+    # been passed as an argument and the CMakeCache exists.
```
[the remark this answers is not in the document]
vigejolla: So what exactly are you suggesting?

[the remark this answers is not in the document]
martyone: I would first consider disallowing to omit the source path on subsequent `mb2 cmake` calls, and if this turns not to be acceptable then consider extending `mb2 qmake` (and `mb2 build`?) so that it also allows to omit the source path on subsequent calls.

sdk-setup/src/mb2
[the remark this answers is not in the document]
martyone: IIUC those are the arguments Qt Creator passes. What I am not sure about is should we run cmake directly, i.e. bypass rpmbuild, in other cases? Is there anything special about the other cases? I would expect that cmake is always executed with the help of rpmbuild as it is done with qmake (except the case when there is no .spec which is consider a legacy option that would be better dropped already.)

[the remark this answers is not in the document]
vigejolla: I can't of course be certain of Jare's thinking, but I would guess the idea was the the developers can run cmake manually themselves, passing whatever arguments they want.

sdk-setup/src/mb2
```diff
@@ -1389,6 +1448,83 @@ run_prepare() {
...
+    local rpmbuild="--build . --target all"
+    local sb2args=( -t "$OPT_TARGET" env PATH="${WRAPPERSDIR}:${USERWRAPPERSDIR}:${PATH}" )
+
+    # CMake projects need to have dependencies installed before running
+    # any CMake commands
+    if [[ $OPT_SPEC ]]; then
+        verify_target_dependencies
+    fi
+
+    if [[ " $* " == *"$rpmbuild"* && $OPT_SPEC ]]; then
```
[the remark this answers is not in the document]
vigejolla: done

sdk-setup/src/mb2
```diff
@@ -1389,6 +1448,83 @@ run_prepare() {
...
+
+        cp -a "$OPT_SPEC" "$TMP_SPEC"
+
+        # get rid of potential windows linefeeds
+        sed -i "s/\x0D$//g" "$TMP_SPEC"
+
+        fix_package_version "$TMP_SPEC"
+
+        eval "set -- $(printf "%q " "$@" |sed "s/ $rpmbuild / /")"
+        init_cmake_wrapper $WRAPPERSDIR/cmake "$OPT_SRC_DIR"
```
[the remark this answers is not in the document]
vigejolla: fixed

sdk-setup/src/mb2
```diff
@@ -2528,8 +2683,18 @@ fi
...
 
+if [[ -f "CMakeCache.txt" ]]; then
+    cmake_pre_setup
+fi
+
 $cmd "$@"
 
+if [[ -f "CMakeCache.txt"
+        && -n "$MER_SSH_HOST_SHARED_HOME"
+        && -n "$MER_SSH_HOST_SHARED_SRC" ]]; then
```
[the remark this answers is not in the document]
vigejolla: done

sdk-setup/src/mb2
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 #
-# Copyright (C) 2013-2019 Jolla Ltd.
-# Copyright (C) 2019-2020 Open Mobile Platform LLC.
+# Copyright (C) 2013 - 2020 Jolla Ltd.
+# Copyright (C) 2019 - 2020 Open Mobile Platform LLC.
```
martyone, line 4: hm, isn't it more common to omit spaces here?

sdk-setup/src/mb2
```diff
@@ -1389,6 +1450,81 @@ run_prepare() {
...
+            echo sb2 "${sb2args[@]}" rpmbuild --build-in-place --target=$build_tgt --define "noecho 1" \
+                --define "_sourcedir $(dirname $OPT_SPEC)" -bc --short-circuit "$TMP_SPEC"
+            
+            sb2 "${sb2args[@]}" rpmbuild --build-in-place --target=$build_tgt --define "noecho 1" \
+                --define "_sourcedir $(dirname $OPT_SPEC)" -bc --short-circuit "$TMP_SPEC"
+    
+        )
+    else
+        init_cmake_wrapper $WRAPPERSDIR/cmake "$@"
+        sb2 "${sb2args[@]}" cmake
```
martyone, line 1522: Is the wrapper really needed here? Maybe the following is enough as it is in the qmake case:

`sb2 -t $OPT_TARGET cmake "$@" ${OPT_PRO_FILE_OR_DIR:+"$OPT_PRO_FILE_OR_DIR"}`

In any case the `-t` option is missing.

sdk-setup/src/mb2
```diff
@@ -1389,6 +1450,81 @@ run_prepare() {
...
+    fi
+    local rc=$?
+    return $rc
```
martyone, line 1525: I would drop all occurrences of this. It is likely there was some need to this before but now all occurrences seems to be without effect.

sdk-setup/src/mb2
```diff
@@ -1389,6 +1450,81 @@ run_prepare() {
...
 }
 
+cmake_post_cleanup() {
+    # Handle path changes for Qt Creator
+    local cmakefiles="CMakeCache.txt"
+
+    cp -p CMakeCache.txt CMakeCache.txt.qtccache
+
+    for cbpfile in *.cbp; do
+        [ -f "$cbpfile" ] || continue
```
martyone, line 1460: Should use `[[` and without quotes.

sdk-setup/src/mb2
```diff
@@ -1389,6 +1450,81 @@ run_prepare() {
...
+        cp -p "$cbpfile" "$cbpfile.qtccache"
+        cmakefiles="$cmakefiles $cbpfile"
+    done
+
+    if [[ "$STATEDIR" == "/home/mersdk/share"* && $MER_SSH_HOST_SHARED_HOME ]]; then
```
martyone, line 1465: No need to quote `$STATEDIR`

sdk-setup/src/mb2
```diff
@@ -2023,6 +2160,8 @@ if [[ $1 != --self-test ]]; then  ###  M A I N  EXECUTION BEGINS HERE  #########
 # If this is called with an empty argument, the default settings is restored.
 handle_pro_file_or_dir_option() {
     local pro_file_or_dir=$1
+    local cmakecache_file="CMakeCache.txt"
+    local read_from_cmakecache=
```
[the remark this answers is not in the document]
vigejolla: yup

sdk-setup/src/mb2
```diff
@@ -1389,6 +1450,81 @@ run_prepare() {
...
+
+        eval "set -- $(printf "%q " "$@" |sed "s/ $rpmbuild / /")"
+        init_cmake_wrapper $WRAPPERSDIR/cmake "$@" "$OPT_SRC_DIR"
+        init_simple_wrapper $WRAPPERSDIR/make
+
+        (
+            cd_to_spec_setup_dir
+            
+            echo sb2 "${sb2args[@]}" rpmbuild --build-in-place --target=$build_tgt --define "noecho 1" \
+                --define "_sourcedir $(dirname $OPT_SPEC)" -bc --short-circuit "$TMP_SPEC"
```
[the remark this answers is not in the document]
vigejolla: dropped

sdk-setup/src/mb2
vigejolla: The examples in sage's presentation have spaces

sdk-setup/src/mb2
vigejolla: I think you are right.

sdk-setup/src/mb2
vigejolla: To me it looks like the idea has been to have mb2 exit with the exit status of the executed command. Now this is broken because we do cmake_post_cleanup after the command. Perhaps I should just fix it instead of dropping?

sdk-setup/src/mb2
vigejolla: fixed

sdk-setup/src/mb2
vigejolla: fixed

sdk-setup/src/mb2
martyone: The last *executed* command's exit code defines the exit code of the function, it is not necessary to capture and return it explicitly. I guess there were some additional actions between `local rc=$?` and `return $rc` before, which are gone now, so the extra step is not needed anymore.

You are right cmake_post_cleanup after the command needs to be fixed.

> Perhaps I should just fix it instead of dropping?

So I would recommend both. Current occurrences dropped and one added near the cmake_post_cleanup call.

sdk-setup/src/mb2
```diff
@@ -2528,10 +2646,20 @@ fi
...
 
+rc=$?
+
+if [[ -f "CMakeCache.txt" ]]; then
+    cmake_post_cleanup
+fi
+
 ##############################################################################
-exit; fi ###  S E L F - T E S T  EXECUTION BEGINS HERE #######################
+exit $rc; fi ###  S E L F - T E S T  EXECUTION BEGINS HERE ###################
```
martyone, line 2662: I would separate the exit on its own line now.

sdk-setup/src/mb2
```diff
@@ -1385,8 +1444,75 @@ run_prepare() {
...
+        eval "set -- $(printf "%q " "$@" |sed "s/ $rpmbuild / /")"
+        init_cmake_wrapper $WRAPPERSDIR/cmake "$@" "$OPT_SRC_DIR"
+        init_simple_wrapper $WRAPPERSDIR/make
+
+        (
+            cd_to_spec_setup_dir
+            
+            sb2 "${sb2args[@]}" rpmbuild --build-in-place --target=$build_tgt --define "noecho 1" \
+                --define "_sourcedir $(dirname $OPT_SPEC)" -bc --short-circuit "$TMP_SPEC"
+
```
martyone, line 1511: extra blank line

sdk-setup/src/mb2
```diff
@@ -1385,8 +1444,75 @@ run_prepare() {
...
+        mv CMakeCache.txt.qtccache CMakeCache.txt
+        for cbp_file in *.cbp.qtccache; do
+            mv $cbp_file ${cbp_file%.qtccache}
+        done
+    fi
+}
+
+run_cmake() {
+    local rpmbuild="--build . --target all"
+    local sb2args=( -t "$OPT_TARGET" env PATH="${WRAPPERSDIR}:${USERWRAPPERSDIR}:${PATH}" )
```
martyone, line 1483: I would use those literally as it is done under run_qmake

sdk-setup/src/mb2
```diff
@@ -1194,6 +1200,60 @@ END
...
+
+        if [[ $IS_SHADOW_BUILD ]]; then
+            subdir=${PWD#$BUILD_DIR}
+            src_dir=$SRC_DIR${subdir:+/$subdir}
+            args=()
+            for arg in "$@"; do
+                args+=("$arg")
+            done
+            args+=("$src_dir")
+            set -- "${args[@]}"
```
martyone, line 1232: Compared to qmake case we do not do any $arg manipulation and append $src_dir unconditionally here. Is this really/still correct? If so, then I would simply replace this with `set -- "$@" "$src_dir"`.

sdk-setup/src/mb2
vigejolla: fixed

sdk-setup/src/mb2
vigejolla: removed

sdk-setup/src/mb2
vigejolla: fixed

sdk-setup/src/mb2
```diff
@@ -244,9 +243,31 @@ COMMANDS
         project directory. (The file name part is ignored - project file
         selection is done by qmake call inside the .spec file.)
 
+    cmake [<project-dir>] [[--] <cmake-extra-args>]
```
vigejolla, line 246: cmake requires that the project dir is the last argument. I think we should keep the same requirement. So the format for cmake invokations should be "cmake <cmake-extra-args> <project-dir>".

martyone: mb2 needs to restrict it to either first or last argument to be able to recognize it reliably between all the possible arguments (there may be options accepting paths as arguments). cmake actually does not insist on argument order in this case, and qtc invokes it exactly like that - project dir first, options later. That's why mb2 requires it to be the very first argument. The same applies to the qmake command.

sdk-setup/src/mb2
```diff
@@ -78,9 +76,9 @@ the "build" or "qmake" command:
 
 Limitations related to shadow builds:
 
-    1. Shadow build is only available for packages that use qmake and
-       their build and install procedure is fully realized using qmake
-       with no additional steps implemented at .spec level.
+    1. Shadow build is only available for packages that use qmake (or CMake)
+       and their build and install procedure is fully realized using qmake (or
+       CMake) with no additional steps implemented at .spec level.
```
vigejolla, line 81: I would drop the parenthesis and break the long sentence:
`1. Shadow build is only available for packages that use qmake or CMake. Their build and install procedure must be fully realized using qmake or CMake, with no additional steps implemented at .spec level.`

martyone: In future there will be also autotools and possibly others, so I think it is better to formulate the common parts for the "preferred" option and only indicate the alternate options this way.

vigejolla: We will have to modify the text anyway at that point

martyone: I mean at that point we would just extend it but not restructure, e.g.,

 "...that uses qmake (or CMake or GNU Autotools) and their..."

and where possible (as in the latter case here) just some shortened form like

 "...realized using qmake (CMake etc.) with..."

vigejolla: I get what you mean, but I think we should anyway try to improve the readability of the text. Having long sentences and more and more text inside parenthesis does not exactly make it easy to read and comprehend. It's not (yet) a disaster, but anyway it has room for improvement.

martyone: I am worried about that too. I was able to split the sentence after all.

vigejolla: Looks good, thanks :)

PR #149, opened 2020-01-22
Jb48360

jlehtoranta:
Only the last commit belongs to this PR


martyone:
Turned to not be needed after all (at least not now)

PR #151, opened 2020-01-29
Jb45577

martyone:


martyone:
Just the last two commits belong to this PR

sdk-setup/src/mb2
```diff
@@ -320,9 +320,10 @@ COMMANDS
             root access to the device.
 
         --rsync
-            Uses 'rsync' to copy the installed files under '/opt/sdk/<name>'
-            without building/installing RPM package(s).  Does not require root
-            access to the device.  Incompatible with '--all'.
+            This method is special.  Uses 'rsync' to copy the installed files
```
vigejolla: I don't think "This method is special" is useful information. Instead, perhaps start by stating "This method does not deal with RPM packages. Instead it ..."

sdk-setup/src/mb2
```diff
@@ -341,6 +342,66 @@ COMMANDS
...
+            Uses 'pkcon' to remove RPM package(s).  Requires the "Allow
+            untrusted software" option enabled in Settings, does not require
+            root access to the device.
+
+        --rpm
+            Uses plain 'rpm' to remove RPM package(s).  Requires root access
+            to the device.
+
+        --rsync
+            This method is special.  It does not work with RPM packages.  It
```
vigejolla: I think the mention about being special is not interesting here as well. "This method does not work with RPM packages" would be sufficient.

sdk-setup/src/mb2
```diff
@@ -1027,14 +1027,14 @@ get_device() {
 rsync_as() {
     local user="$1";shift
     local key=$(get_shared_dir)/${device_sshkeypath}/$user
-    [[ -f "$key" ]] || fatal "No key for user '$user' on '$device_name' given in devices.xml (rsync '$key')"
+    [[ -f "$key" ]] || fatal "SSH key not configured for user '$user' on '$device_name'"
```
vigejolla: Shouldn't we give the user some hint about how to configure the key?

sdk-setup/src/mb2
martyone: I want to clearly distinguish this method from the others. The fact that it does not deal with RPM packages has other implications. Only all these together make this method so special, and this is something one may not easily realize. Hence the tl;dl message. Also, the main description only deals with this method as little as possible - otherwise if would be full of "except for" and "only if".

Now I noticed I left "The '--all' option is not available when the '--rsync' method is used." in the main description - this is to be moved here too.

sdk-setup/src/mb2
martyone: ditto

sdk-setup/src/mb2
martyone: I though about that too, but didn't like it. Not only that specifically in case of mb2 there are three frontedns (including mb2 itself), each deserving different handling. In general hints hardcoded in code are not ideal. It is better to issue a brief but distinguishable error message and have an answer elsewhere easy to look up based on the error message (docs, TJC, ...) - such an information is much easier to expand, improve, correct, ..., without all the integration hassle.

PR #152, opened 2020-02-05
[mb2] Fix use with %autosetup -pN. Fixes JB#47988

martyone:


martyone:
Also the branch name is wrong, a new PR is needed :(

PR #156, opened 2020-03-02
mb2: Log package install changes.

jusa:
If new packages need to be installed or existing packages are updated
echo the zypper output for the actions.


martyone:
Only now after merging #225 I noticed your original PR in my notes. I thought about limiting the output as you did, but it would break with other than the default locale used and it could also hide useful information in the preceding output, so it would need to be checked more. Please file a bug if you feel this really needs to be handled.

martyone:
Superseded by #225

PR #157, opened 2020-03-04
Jb45810

martyone:


sdk-setup/src/oomadvice
```diff
@@ -0,0 +1,199 @@
...
+{
+    local fn=$1
+    local args=("${@:2}")
+
+    bash -c "$(declare -f $fn); $fn \"\$@\"" bash "${args[@]}"
+}
+
+set_defaults()
+{
+    OPT_OOM_ADJ=10
```
vigejolla: How did you come up with the value 10?

martyone: Just some positive value.

martyone:
Rebased, updated conditions for running under vbox only

vigejolla:
Still LGTM

PR #162, opened 2020-03-12
[mb2] Fix reading ugly formatted devices.xml. JB#48741

martyone:


vigejolla:
JB#48741 has been closed almost a month ago. This probably deserves an own bug.

PR #164, opened 2020-04-09
Jb49476

martyone:


martyone:
Just the very last commit belongs to this PR

martyone:
> I suppose we can do it like this for this release. But we should make the real fix directly to mer-tooling-chroot.

Not really, the old toolings need to be fixed too, so some kind of postprocessing is inevitable.

vigejolla:
> > I suppose we can do it like this for this release. But we should make the real fix directly to mer-tooling-chroot.
> 
> Not really, the old toolings need to be fixed too, so some kind of postprocessing is inevitable.

While that's true, I still think that it would be better if we had the fix already in mer-tooling-chroot. It's not fun to maintain scripts which are postprocessed by scripts here and there. Our postprocessing script just needs to be smart enough not to apply the fix if it already exists in mer-tooling-chroot. Besides, I doubt mer-tooling-chroot will never be changed in future. Our postprocessing here might just stop working with the next release.

martyone:
Yup, I didn't mean it that I am against fixing future toolings :)

martyone:
> I'm just thinking if there was a way to test if --times would actually work. But I guess just disabling it always on docker is OK as well.

hm, IIUC including or not including --times can significantly affect execution time, so maybe better to keep it under explicit control?

vigejolla:
> hm, IIUC including or not including --times can significantly affect execution time, so maybe better to keep it under explicit control?

Is that so? In that case I agree that it's probably better to keep it explicitly disabled.

martyone:
Replaced with https://github.com/sailfishos/sdk-setup/pull/165 under different bz item

PR #169, opened 2020-05-11
Jb49843

martyone:


martyone:
> So, caching CMake version was needed after all. LGTM.

Yeah I am not sure why I didn't notice it complaining about before. Maybe because the error is not fatal (strange).

PR #172, opened 2020-05-18
Jb45811

martyone:


sdk-setup/src/xmp.py
```diff
@@ -0,0 +1,278 @@
+#!/usr/bin/env python
+
+#    Copyright (C) 2001  Jeff Epler  <jepler@unpythonic.dhs.org>
+#    Copyright (C) 2006  Csaba Henk  <csaba.henk@creo.hu>
+#
+#    This program can be distributed under the terms of the GNU LGPL.
+#    See the file COPYING.
```
vigejolla: I don't see any COPYING file here - please add it

sdk-setup/src/dynexecfs
```diff
@@ -1,7 +1,8 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 #    Copyright (C) 2001  Jeff Epler  <jepler@unpythonic.dhs.org>
 #    Copyright (C) 2006  Csaba Henk  <csaba.henk@creo.hu>
+#    Copyright (C) 2019  Open Mobile Platform LLC.
```
vigejolla, line 5: It's 2020 already

sdk-setup/src/xmp.py
martyone: This is the vanilla source imported from python-fuse.

sdk-setup/src/dynexecfs
martyone: done

martyone:
Time to add `:abbr pyhton python` :)

sdk-setup/src/xmp.py
vigejolla: I know. The COPYING file in python-fuse contains the GNU LGPL text, which states: "distribute a copy of this License along with the
Library". Which means that we *must* include also the COPYING file.

martyone: ah, now I got it, I read too fast and misread :) I added reference to gnu.org/licenses instead, how do you like it?

vigejolla: Unfortunately it does not matter if I like it or not - the license specifically states that we must supply a copy of the license. It's a bit funny, as the license also states that it is possible to use any version of the license, so in my mind it would make more sense to point to gnu.org. But I guess they want to make sure that the license text is still readable even if gnu.org goes down for some reason.

martyone: Initially I thought that too but a quick check didn't reveal such a requirement. Unless you found it I would go with the link to GNU site as a more convenient option in this particular case.

vigejolla: https://github.com/libfuse/python-fuse/blob/master/COPYING#L155 
https://github.com/libfuse/python-fuse/blob/master/COPYING#L280

martyone: pity :\ check now please.

PR #173, opened 2020-05-19
[mb2] Treat 106 as successful return value for zypper. Fixes JB#46187

vigejolla:
zypper returns 106 when it has disabled some repository temporarily.
It does not mean that the installation had been unsuccessful.


vigejolla:
> Is this the only non-zero code to consider? It would be nice if there was an option or environment variable that would prevent zypper reporting non-fatal issues this way.

Looking at the documentation, I believe that all other error codes should be really considered as errors. I agree that it would be nice to have an option to change the behaviour, but I couldn't find such.

PR #174, opened 2020-05-28
Jb49038

martyone:


sdk-setup/src/sdk-manage
```diff
@@ -1063,9 +1121,23 @@ manage_object_packages() {
     fi
 
     object_zypper "$type:$object" --no-refresh "$action" "${packages[@]}" || return
-    if [[ $INSIDE_VBOX && $type == target ]] && ! target_is_snapshot "$object"; then
-        synchronise_target "$object"
+    maybe_synchronize_object "$type:$object"
+}
+
+maybe_synchronize_object() {
```
vigejolla, line 1127: Since we are already using the British spelling in synchronise_target, could we use it also here (i.e. maybe_synchronise_object)?

martyone: yup :)

rpm/sdk-setup.spec
```diff
@@ -85,6 +85,7 @@ Requires:   git
 Requires:   tar
 Requires:   p7zip-full
 Requires:   gnu-cpio
+Requires:   compiledb
```
vigejolla, line 88: Just checking that I understand correctly this change: Instead of running compiledb on the targets, we now run only make (in dry-run mode) in target, and parse the output on build engine side?

vigejolla: ... reading the commit message would of course help ;) LGTM

PR #177, opened 2020-06-02
[mb2] Use target name as qmake variable cache filename infix. JB#49115

vigejolla:


vigejolla:
> I would consider using it as infix instead, keeping "cache" as the file extension.

Sure, I can see how that would make sense

PR #178, opened 2020-06-03
Jb45818

martyone:


sdk-setup/src/mb2
```diff
@@ -3072,9 +3072,7 @@ if [[ ! "$OPT_SPEC" ]]; then
     spec_files=("$OPT_PKGDIR"/*.spec)
     numspec=${#spec_files[@]}
     if [[ $numspec -gt 1 ]]; then
-        [[ $OPT_NEEDSPEC ]] && fatal "Multiple spec files found - use --specfile to select one"
-        echo "Multiple spec files found - not using any. Use --specfile to select one"
-        # OPT_SPEC is not set
+        fatal "Multiple spec files found - please select one."
```
vigejolla, line 3075: I don't get it - to me it seems the old version should have worked just as well. Now we just refuse to execute even those commands which don't need a spec file if there are multiple spec files. What's the point?

martyone: Yeah, not ideal too. The problem is that qmake and make commands work without .spec. I believe this was implemented to allow opening/building projects in Qt Creator before .spec is created. While those are said to not need spec, they clearly should honor a spec if there is at least one, so they should fail if more than one exist and none was selected explicitly. With the old version it would just print the message and continue using no spec :D

I think we have two options
1) let qmake/make strongly require spec
2) add OPT_WANTSPEC in addition to OPT_NEEDSPEC or something like that to add more granularity.

vigejolla: I would be happy with the first option. Sure, there might be someone somewhere who wants to use QtC to open projects before creating a .spec file, but still I doubt it's worth the extra trouble of adding OPT_WANTSPEC.

martyone: Yup, and should we need to revert that, we can do so on merssh side by invoking 'sfdk build-shell qmake' instead of 'sfdk qmake' if there is no spec.

martyone: Updated

PR #179, opened 2020-06-09
Jb50188

etaishev:
Add variant for deploy RPMs without build and calling the sign utility

### DIFF
--- a/sdk-setup/src/mb2
+++ b/sdk-setup/src/mb2
@@ -200,6 +200,35 @@ section in the spec file.  If a file with '.changes.run' extension is
 found instead, this file will be executed and its output treated as
 the actual changelog.
 
+
+SIGNING PACKAGES
+
+The 'build' and 'package' commands allow to optionally sign the resulting
+packages. Use the '--sign' option to these commands to enable this step.
+
+The cryptographic key to use for signing can be selected with the global
+option '--package-signing-user'. If this is a passphrase-protected key, the
+passphrase may be supplied with one of the '--package-signing-passphrase' or
+'--package-signing-passphrase-file' options. Set up yourself a shell alias to
+apply these by default:
+
+    $ touch PASSFILE
+    $ chmod 600 PASSFILE
+    $ cat >PASSFILE
+    YourPassphrase
+    ^D
+    $ alias mb2="mb2 --package-signing-user NAME \\
+        --package-signing-passphrase-file PASSFILE"
+
+The GPG version on host and under the SDK are not likely to match - for that
+reason the SDK uses its own GPG home directory and therefore your secret key
+must be imported under the SDK before it can be used. This can be achieved
+with command like this (executed on host):
+
+    $ gpg --export-secret-keys ID |sdk gpg2 --import
+
+(Replace 'sdk' with the command you use to enter the SDK)
+
 COMMANDS
     build-requires [--[no-]refresh] {pull|reset}
         When 'pull' is used, install or update the build-time dependencies as
@@ -225,7 +254,7 @@ COMMANDS
 
         -R    reverse apply patches (patch -R)
 
-    build [-p|--prepare] [--no-check] [-d|--enable-debug] [-j <n>]
+    build [-p|--prepare] [--no-check] [-d|--enable-debug] [-j <n>] [-s|--sign] 
           [<project-dir-or-file>] [[--] <rpmbuild-extra-arg>...]
         Run rpmbuild. Execute all sections of the .spec file except for %prep
         unless told otherwise.
@@ -239,6 +268,7 @@ COMMANDS
         --no-check        Skip the %check section
         --enable-debug    Enable debug build
         -j <n>            Run up to <n> jobs simultaneously
+        --sign            Sign packages
 
     qmake [<project-dir-or-file>] [[--] <qmake-extra-arg>...]
         Run rpmbuild. Execute just the %build section of the .spec file,
@@ -283,10 +313,10 @@ COMMANDS
     make-install
         Run rpmbuild. Execute just the %install section of the .spec file.
 
-    package [--no-check]
+    package [--no-check] [-s|--sign]
         Run rpmbuild. Execute just the %install section of the .spec file, then
-        build binary packages. Unless '--no-check' is passed, execute also the
-        %check section if present.
+        build binary packages and optionally sign those. Unless '--no-check'
+        is passed, execute also the %check section if present.
 
     build-shell [<command> [<arg>...]]
         Execute an arbitrary <command> in the build environment.  If no
@@ -601,6 +631,22 @@ GLOBAL OPTIONS
     --check-suites [+|-]<suite>...
         Override the default test suites the 'check' command executes. See
         the '--suites' option to the 'check' command for more details.
+
+    --package-signing-user <name>
+        Select the GPG key to sign packages with. Accepts the same values
+        as the '--local-user' option of the gpg command does. See GPG
+        documentation for more details.
+
+        Depending on your SDK variant, different semantics may apply.
+
+    --package-signing-passphrase <string>
+        In case a passphrase-protected signing key is selected for signing
+        packages, use <string> as the passphrase.
+        
+    --package-signing-passphrase-file <file>
+        In case a passphrase-protected signing key is selected for signing
+        packages, read the passphrase from the <file>. Only the first line will
+        be read.
 
 
 KNOWN ISSUES
@@ -1809,6 +1855,106 @@ run_build_requires__process_args() {
     fi
 }
 
+sign_rpms() {
+    [[ -e $STATEDIR/spec ]] || fatal "No previous build found"
+    local rpms=$(find_rpms $all_opt "$OPT_OUTPUTDIR" "$STATEDIR/spec")
+    [[ $rpms ]] || fatal "No package found"
+
+    local sdksign=/usr/libexec/sdk-setup/sdk-rpmsign
+
+    if [[ -x $sdksign ]]; then
+        $sdksign ${rpms}
+    else
+        sign_rpms_native ${rpms}
+    fi
+
+    local rc=$?
+    [[ $rc -eq 0 ]] || { notice "Failed to sign packages"; return $rc; }
+}
+
+sign_rpms_native() {
+    local rpms=("$@")
+
+    local tmpdir=
+    sign_rpms_native_cleanup() (
+        trap 'echo cleaning up...' INT TERM HUP
+        if [[ $tmpdir ]]; then
+            rm -rf "$tmpdir"
+        fi
+    )
+    trap 'sign_rpms_native_cleanup; trap - RETURN' RETURN
+    trap 'return 1' INT TERM HUP
+
+    local extra_args=()
+
+    if [[ $OPT_PACKAGE_SIGN_USER ]]; then
+
+        if inside_build_engine; then
+            local shared_key=${OPT_SHARED_DIR}/gnupg/${OPT_PACKAGE_SIGN_USER}.key
+            if [[ ! -f $shared_key ]]; then
+                notice "Cannot sign packages: Internal error: No GPG key shared for the user '$OPT_PACKAGE_SIGN_USER'."
+                return 1
+            fi
+
+            local shared_key_id=$(gpg2 --with-colons --list-packets "$shared_key" \
+                    |sed -n 's/:user ID packet: "\(.*\)"/\1/p')
+            [[ $shared_key_id ]] || return
+
+            local key_fingerprints=$(gpg2 --with-colons --list-secret-keys --fingerprint "$shared_key_id" \
+                    |awk -F: '$1 == "fpr" { print $10 }')
+            local key_fingerprint=
+            for key_fingerprint in $key_fingerprints; do
+                if ! gpg2 --batch --delete-secret-and-public-key "$key_fingerprint"; then
+                    notice "Cannot sign packages: Internal error: Failed to delete old GPG key copy '$key_fingerprint'."
+                    return 1
+                fi
+            done
+
+            if ! gpg2 --quiet --import "${shared_key}"; then 
+                notice "Cannot sign packages: Internal error: Failed to import GPG key from file '$shared_key'."
+                return 1
+            fi
+        fi
+
+        if ! gpg2 --list-secret-keys "$OPT_PACKAGE_SIGN_USER" &>/dev/null; then
+            notice "Cannot sign packages: No GPG key found for the user '$OPT_PACKAGE_SIGN_USER'."
+            return 1
+        fi
+
+        extra_args+=(--define "_gpg_name $OPT_PACKAGE_SIGN_USER")
+
+        local passphrase=
+
+        if [[ $OPT_PACKAGE_SIGN_PASSPHRASE_FILE ]]; then
+            passphrase=$(<"$OPT_PACKAGE_SIGN_PASSPHRASE_FILE")
+        else
+            passphrase=$OPT_PACKAGE_SIGN_PASSPHRASE
+        fi
+
+        if ! gpg2 --batch --status-fd 1 --sign --local-user "$OPT_PACKAGE_SIGN_USER" \
+                --output - --passphrase-fd 3 3<<<"$passphrase" </dev/null &>/dev/null; then
+            if [[ ! $passphrase ]]; then
+                notice "Cannot sign packages: The selected GPG key is passphrase protected" \
+                    "and no passphrase was specified."
+            else
+                notice "Cannot sign packages: The given passphrase is not a correct passphrase" \
+                    "for the '$OPT_PACKAGE_SIGN_USER' key."
+            fi
+            return 1
+        fi
+
+        if [[ $passphrase ]]; then
+            tmpdir=$(mktemp -d) || return
+            chmod 0700 "$tmpdir" || return
+            local passphrase_file=$tmpdir/passphrase
+            cat >"$passphrase_file" <<<"$passphrase" || return
+            extra_args+=(--define "_gpg_sign_cmd_extra_args --batch --passphrase-file '$passphrase_file'")
+        fi
+    fi
+
+    rpmsign --addsign "${rpms[@]}" "${extra_args[@]}" </dev/null
+}
+
 run_build_requires() {
     if [[ $BUILD_REQUIRES_MODE == reset ]]; then
         if ! sdk-manage target snapshot --reset=force "$OPT_ORIGINAL_TARGET" "$OPT_TARGET"; then
@@ -1880,6 +2026,7 @@ run_build__process_args() {
     BUILD_NOPREP_ARGS=(--noprep)
     BUILD_NOCHECK_ARGS=()
     BUILD_EXTRA_ARGS=()
+    BUILD_SIGN=
     while [[ "$1" ]]; do
         case "$1" in
             -d|--enable-debug) shift
@@ -1900,6 +2047,9 @@ run_build__process_args() {
                 fi
                 shift
                 BUILD_NOPREP_ARGS=()
+                ;;
+            -s|--sign) shift
+                BUILD_SIGN=1
                 ;;
             --no-check) shift
                 BUILD_NOCHECK_ARGS=(--nocheck)
@@ -1970,8 +2120,10 @@ run_build() {
             "${BUILD_NOPREP_ARGS[@]}" "${BUILD_NOCHECK_ARGS[@]}" \
             -bb \
             "$TMP_SPEC" \
-            "${BUILD_EXTRA_ARGS[@]}"
+            "${BUILD_EXTRA_ARGS[@]}" \
+            || return
     )
+    [[ ! $BUILD_SIGN ]] || sign_rpms
 }
 
 run_prepare() {
@@ -2259,11 +2411,15 @@ run_make_install() {
 
 run_package__process_args() {
     PACKAGE_NOCHECK_ARGS=()
+    PACKAGE_SIGN=
 
     while [[ $# -ne 0 ]]; do
         case $1 in
             --no-check)
                 PACKAGE_NOCHECK_ARGS=(--nocheck)
+                ;;
+            -s|--sign) shift
+                PACKAGE_SIGN=1
                 ;;
             *)
                 fatal "Unexpected argument: $1"
@@ -2298,8 +2454,10 @@ run_package() {
             --define "_buildhost $(get_hostname)" \
             --buildroot="$buildroot" \
             -bb --noprep --nobuildstage "${PACKAGE_NOCHECK_ARGS[@]}" --noclean \
-            "$TMP_SPEC"
+            "$TMP_SPEC" \
+            || return
     )
+    [[ ! $PACKAGE_SIGN ]] || sign_rpms
 }
 
 run_build_shell() {
@@ -2869,6 +3027,9 @@ OPT_INC_BUILD_NUMBER=0
 OPT_PULL_BUILD_REQUIRES=1
 OPT_WRAP=()
 OPT_CHECK_LEVELS=
+OPT_PACKAGE_SIGN_USER=
+OPT_PACKAGE_SIGN_PASSPHRASE=
+OPT_PACKAGE_SIGN_PASSPHRASE_FILE=
 OPT_CHECK_SUITES=
 
 STATEDIR=$PWD/.mb2
@@ -3100,6 +3261,15 @@ while [[ "$1" ]]; do
         "-m" | "--submodule" )
             notice "The '$1' option is deprecated and does nothing currently"
             shift 2
+            ;;
+        "--package-signing-user") shift
+            OPT_PACKAGE_SIGN_USER="$1"; shift
+            ;;
+        "--package-signing-passphrase") shift
+            OPT_PACKAGE_SIGN_PASSPHRASE="$1"; shift
+            ;;
+        "--package-signing-passphrase-file") shift
+            OPT_PACKAGE_SIGN_PASSPHRASE_FILE="$1"; shift
             ;;
         "--check-levels" )
             [[ $2 ]] || fatal "The '$1' option requires an argument"
--- a/sdk-setup/src/mb2
+++ b/sdk-setup/src/mb2
@@ -2511,10 +2511,6 @@ run_deploy() {
 
     # rsync method is special
     if [[ $method_opt == --rsync ]]; then
-        if behind_qtcreator; then
-            run_make_install || return
-        fi
-
         local name=$(get_spec_tag "%{name}")
 
         if [[ $dry_run_opt ]]; then
@@ -2525,10 +2521,6 @@ run_deploy() {
         rsync_as $device_username -av ${buildroot}/. $device_ip:/opt/sdk/$name \
             || { echo "Deploy failed"; return 1; }
         return
-    fi
-
-    if behind_qtcreator; then
-        run_package || return
     fi
 
     local rpms=$(find_rpms $all_opt "$OPT_OUTPUTDIR" "$STATEDIR/spec" "${patterns[@]}")
--- a/sdk-setup/src/mer-sdk-chroot
+++ b/sdk-setup/src/mer-sdk-chroot
@@ -60,6 +60,7 @@ EOF
     return 0
 }
 
+MY_GNUPGHOME_DIR='.config/SailfishSDK/gnupg'
 MY_SSH_AUTH_SOCK=${SSH_AUTH_SOCK#/parentroot}
 [[ $MY_SSH_AUTH_SOCK ]] && MY_SSH_AUTH_SOCK="/parentroot$MY_SSH_AUTH_SOCK"
 
@@ -288,7 +289,14 @@ prepare_user() {
 	# Now the sdk uses a private namespace, there's no need to
 	# make it unbindable
 	mount --make-shared ${sdkroot}${HOMEDIR}
+        # Prevent accidental writes to host's GPG home (SDK uses its own GPG home)
+        if [[ -d ${sdkroot}${HOMEDIR}/.gnupg ]]; then
+            mount --bind -o ro "${sdkroot}${HOMEDIR}/.gnupg"{,}
+        fi
     fi
+
+    sudo -u $user mkdir -p --mode 700 "${sdkroot}${HOMEDIR}/${MY_GNUPGHOME_DIR}"
+
     echo "$user ALL=NOPASSWD: ALL" > ${sdkroot}/etc/sudoers.d/$user
     chmod 0440 ${sdkroot}/etc/sudoers.d/$user
 }
@@ -396,6 +404,7 @@ setarch i386 chroot ${sdkroot} /bin/su -s /bin/bash -l $user -- -c "$(cat <<END
 export SSH_AUTH_SOCK='${MY_SSH_AUTH_SOCK}'
 export SSH_AGENT_PID=${SSH_AGENT_PID:-}
 export MERSDK='$sdkroot'
+export GNUPGHOME='${HOMEDIR}/${MY_GNUPGHOME_DIR}'
 
 sudo oneshot
 


sdk-setup/src/mb2
```diff
@@ -401,6 +401,8 @@ COMMANDS
         and "package" commands unless the '--all' option or '--rsync' method is
         used.
 
+        When the '--deploy-only' option set, only installation start, whitout rpm build.
```
vigejolla, line 407: s/whitout/without

sdk-setup/src/mb2
```diff
@@ -2074,6 +2077,17 @@ run_package() {
...
 }
 
+run_sign() {
+    [[ -e $STATEDIR/spec ]] || fatal "No previoud build found"
+    local retcode=0
+    rpms=$(find_rpms $all_opt "$OPT_OUTPUTDIR" "$STATEDIR/spec")
+    [[ $rpms ]] || fatal "No package found"
+    local method;
+    method="custom-rpm-sign"
+    $method "$@" ${rpms}
```
vigejolla, line 2087: Are you planning to have an option to change the method? If not, then I suggest you just use custom-rpm-sign directly without the variable.

vigejolla:
Looks already quite good, but we can't really merge this before we have implemented custom-rpm-sign using the standard rpmsign.

sdk-setup/src/mb2
etaishev: Fixed

sdk-setup/src/mb2
etaishev: There is no great need for this now. But it would be convenient for us to use our sign script.

Added a signature script as well

sdk-setup/src/mb2
```diff
@@ -2124,7 +2145,9 @@ run_deploy() {
             else
                 method="pkcon --plain --noninteractive install-local"
             fi
-            ! behind_qtcreator || run_package || return
+            if [[ -z $no_build ]]; then
+                ! behind_qtcreator || run_package || return
+            fi
```
martyone, line 2150: The `run_package` step is only implied when running behind Qt Creator. At the same time your change on QtC side disables that by adding `--deploy-only` to every `mb2 deploy` invocation. So unless I missed something the same effect can be achieved by simply dropping the implied `run_package` execution here.

sdk-setup/src/mb2
```diff
@@ -351,7 +351,10 @@ COMMANDS
         effect is cumulative and persists until overriden with later name
         prefixed with the opposite operator.
 
-    deploy {--pkcon|--rsync|--sdk|--zypper|--manual} [--all] [--debug]
+    sign 
+        Generate and insert new signatures for each collected package, replacing any existing signatures.
```
martyone, line 355: The 'sign' command doesn't do anything special in addition to collecting the related rpms. So, instead of introducing a new command, I would prefer adding '--sign' option to the 'build' and 'package' commands.

sdk-setup/src/customer-sign
```diff
@@ -0,0 +1,23 @@
...
+    echo "  -h, --help            show this help message and exit"
+}
+
+main() {
+    if [[ ( "$1" == "--help") ||  "$1" == "-h" ]]; then 
+        usage
+        exit 0
+    fi 
+
+    rpmsign --addsign "$@"
```
martyone, line 20: If we need to support standard rpmsign, we need to obey its UI, so mb2 may not assume other options being accepted than those the standard rpmsing accepts. For that reason it makes more sense to hardcode use of '--addsign' option in mb2 and only allow to override the rpmsign executable. This could be done e.g. by looking for /usr/libexec/sdk-setup/sdk-rpmsign first and if that is not found continuing with rpmsign from PATH. I.e. no wrapper needs to be provided by default and if provided, mb2 will discover and use it instead of plain rpmsign.

sdk-setup/src/mb2
etaishev: The `run_package` step will be ignored only from `sfdk` if I checked correctly. So `--deploy-only` also need for use insede build engene

sdk-setup/src/mb2
etaishev: There will be no additional build step on the qtcreator side?

sdk-setup/src/mb2
martyone: `! behind_qtcreator || run_package || return` is the same as `if behind_qtcreator; then run_package || return; fi`

sdk-setup/src/mb2
martyone: 'build' and 'package' are the only commands that result in RPM creation (unless I missed something), so only those need the '--sign' option. Hopefully I answered your question as I am not sure I follow.

etaishev: Yes. But what if we want to sign a RPM that has already been created?

martyone: In that case we simply use rpmsign (or its variant) directly.

etaishev: I update PR: add argument for 'build' and 'package'. By default sign is enable, disable it with '--nosign'. This option is the most correct and convenient, but in this case, in PR-https://github.com/sailfishos/sailfish-qtcreator/pull/325 we will lose the sign build step. I wanted to use it for us, to use a custom signature command with its own parameters and output parsing.

sdk-setup/src/mb2
```diff
@@ -225,7 +225,7 @@ COMMANDS
 
         -R    reverse apply patches (patch -R)
 
-    build [-p|--doprep] [-d|--enable-debug] [-j <n>] [<project-dir-or-file>]
+    build [-p|--doprep] [-d|--enable-debug] [-j <n>] [<project-dir-or-file>] [--nosign]
```
martyone, line 228: I am not sure signing should be the default option. It requires some additional setup, we don't have supporting tools for key management, we miss guidelines etc. If we want to save users from typing extra chars, I would prefer addressing that by providing single letter option `[-s|--sign]` now.

The option should be listed before `<project-dir-or-file>`.

sdk-setup/src/mb2
```diff
@@ -351,7 +354,7 @@ COMMANDS
         effect is cumulative and persists until overriden with later name
         prefixed with the opposite operator.
 
-    deploy {--pkcon|--rsync|--sdk|--zypper|--manual} [--all] [--debug]
+    deploy {--pkcon|--rsync|--sdk|--zypper|--manual} [--all] [--debug] [--nosign]
```
martyone, line 357: As we discussed before, it should be changed so that `deploy` never implies `package` (currently it does behind qtc only). When this is done, no signing option (--sign/--nosing) should be needed here.

sdk-setup/src/mb2
```diff
@@ -1775,6 +1779,7 @@ run_build() {
             "$TMP_SPEC" \
             "$@"
     )
+    [[ -z $NO_SIGN_RPM ]] && run_sign
```
martyone, line 1782: The variable should be consistently named BUILD_*

sdk-setup/src/mb2
```diff
@@ -2070,6 +2075,24 @@ run_package() {
             -bb --noprep --nobuildstage --noclean \
             "$TMP_SPEC"
     )
+    [[ -z $NO_SIGN_RPM ]] && run_sign
+}
+
+run_sign() {
```
martyone, line 2081: I would prefer not using run_ prefix here - this is used for functions implementing subcommands. I would call it sign_rpms and move its definition closer to the beginning of the file where other reusable functions are defined.

sdk-setup/src/mb2
```diff
@@ -2070,6 +2075,24 @@ run_package() {
...
+    [[ -e $STATEDIR/spec ]] || fatal "No previoud build found"
+    local retcode=0
+    rpms=$(find_rpms $all_opt "$OPT_OUTPUTDIR" "$STATEDIR/spec")
+    [[ $rpms ]] || fatal "No package found"
+    local sdksign=/usr/bin/sdk-rpmsign
```
martyone, line 2086: This should remain as a kind of configuration detail of mb2. It should not introduce a requirement on an user-oriented tool. So I would prefer using the `/usr/libexec/sdk-setup/sdk-rpmsign` path as I suggested before.

sdk-setup/src/mb2
```diff
@@ -2070,6 +2075,24 @@ run_package() {
...
+
+    if [[ ! -f $sdk-sign ]]; then
+        rpmsign --addsign ${rpms}
+    else
+        $sdk-sign $@ ${rpms}
```
martyone, line 2091: As discussed before either the arguments should be the same in both cases. If we need to support signing tools with different option set, we need to provide more appropriate UI for it.

What would be an example of options not recognized by standard rpmsign?

sdk-setup/src/mb2
```diff
@@ -2841,6 +2867,19 @@ case "$1" in
         cmd=run_${1//-/_}
         shift
         ;;
+    package )
+        cmd=run_$1; shift
+        NO_SIGN_RPM=
```
martyone, line 2872: The variable should be named with PACKAGE_ prefix.

sdk-setup/src/mb2
etaishev: May be i miss something, but how it's work?
In `MerMb2RpmDeployStep` call only mb2 deploy and in each time it rebuild rpm packages.

martyone: Here you are changing the deploy configuration so that it does building and deploying RPM packages with separate steps https://github.com/sailfishos/sailfish-qtcreator/pull/325/commits/557dbb18645bcaaf7e985a8245f5caf3215ba253 .

And here https://github.com/sailfishos/sdk-setup/pull/179#discussion_r446943892 I suggest that instead of achieving that by adding `--deploy-only` we should stop implying run_package with run_deploy.

So since the deploy commands does not rebuild the packages, it does not need the signing option.

sdk-setup/src/mb2
```diff
@@ -1775,6 +1795,7 @@ run_build() {
             "$TMP_SPEC" \
             "$@"
```
martyone, line 1796: Now that it is not the last command in the function, we need to handle possible failure:

```suggestion
            "$@" \
            || return
```

sdk-setup/src/mb2
```diff
@@ -1775,6 +1795,7 @@ run_build() {
...
     )
+    [[ -z $BUILD_SIGN_RPM ]] || sign_rpms
```
martyone, line 1798: Please prefer `!` instead of `-z`.

sdk-setup/src/mb2
```diff
@@ -2070,6 +2091,7 @@ run_package() {
             -bb --noprep --nobuildstage --noclean \
             "$TMP_SPEC"
     )
+    [[ -z $PACKAGE_SIGN_RPM ]] || sign_rpms
```
martyone, line 2094: Ditto - error handling and `-z`.

sdk-setup/src/mb2
```diff
@@ -235,6 +235,8 @@ COMMANDS
         project directory. (The file name part is ignored - project file
         selection is done by qmake call inside the .spec file.)
 
+        If --sign enabled will be generate and insert new signatures for each collected package.
+
```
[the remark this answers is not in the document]
etaishev: Updated PR

sdk-setup/src/mb2
etaishev: If use custom sign tools it may be additional signature keys or parameters

sdk-setup/src/mb2
etaishev: Сhanged

sdk-setup/src/mb2
etaishev: Oh, yes, sorry, updated deploy command according new deploy mechanism

sdk-setup/src/mb2
```diff
@@ -225,7 +225,7 @@ COMMANDS
 
         -R    reverse apply patches (patch -R)
 
-    build [-p|--doprep] [-d|--enable-debug] [-j <n>] [<project-dir-or-file>]
+    build [-p|--doprep] [-d|--enable-debug] [-j <n>] [-s|--sign] [<project-dir-or-file>]
```
martyone, line 228: Looks like a trailing whitespace.

sdk-setup/src/mb2
```diff
@@ -282,9 +283,11 @@ COMMANDS
     make-install
         Run rpmbuild. Execute just the %install section of the .spec file.
 
-    package
+    package [-s|--sign]
         Run rpmbuild. Execute just the %install section of the .spec file, then
-        build binary packages.
+        build binary packages and optionally sign those.
+        
+        -s    sign packages
```
martyone, line 290: I would maybe omit the single option listing here. It's clear enough.

sdk-setup/src/mb2
```diff
@@ -1675,6 +1678,23 @@ switch_to_snapshot() {
...
+    rpms=$(find_rpms $all_opt "$OPT_OUTPUTDIR" "$STATEDIR/spec")
+    [[ $rpms ]] || fatal "No package found"
+    local sdksign=/usr/libexec/sdk-setup/sdk-rpmsign
+
+    if [[ ! -f $sdk-sign ]]; then
+        rpmsign --addsign ${rpms}
+    else
+        $sdk-sign $@ ${rpms}
+    fi
+    retcode=$?
```
martyone, line 1693: retcode is only used here, so better to not declare earlier and `rc` is more common variable name used elsewhere.
```suggestion
    local rc=$?
```

sdk-setup/src/mb2
```diff
@@ -1675,6 +1678,23 @@ switch_to_snapshot() {
...
+
+    [[ $retcode -eq 0 ]] || { echo "Rpm sign failed"; return 1; }
```
martyone, line 1695: The actual return code should be forwarded.

A more appropriate message would be "Failed to sign packages" and it should be printed using the `notice` function instead.

sdk-setup/src/mb2
```diff
@@ -1675,6 +1678,23 @@ switch_to_snapshot() {
...
+    [[ -e $STATEDIR/spec ]] || fatal "No previoud build found"
+    local retcode=0
+    rpms=$(find_rpms $all_opt "$OPT_OUTPUTDIR" "$STATEDIR/spec")
+    [[ $rpms ]] || fatal "No package found"
+    local sdksign=/usr/libexec/sdk-setup/sdk-rpmsign
+
+    if [[ ! -f $sdk-sign ]]; then
+        rpmsign --addsign ${rpms}
+    else
+        $sdk-sign $@ ${rpms}
```
martyone, line 1691: This still needs to be fixed somehow. We are trying to make the actual tool selection hidden to the user and at the same time we want to allow (or even require) the user to pass some tool-specific options here. These decisions are conflicting.

sdk-setup/src/mb2
```diff
@@ -1675,6 +1678,23 @@ switch_to_snapshot() {
...
 }
 
+sign_rpms() {
+    [[ -e $STATEDIR/spec ]] || fatal "No previoud build found"
+    local retcode=0
+    rpms=$(find_rpms $all_opt "$OPT_OUTPUTDIR" "$STATEDIR/spec")
+    [[ $rpms ]] || fatal "No package found"
+    local sdksign=/usr/libexec/sdk-setup/sdk-rpmsign
+
+    if [[ ! -f $sdk-sign ]]; then
```
martyone, line 1688: better test with `-x` as well and possibly omit negation and swap if/else blocks.

sdk-setup/src/mb2
etaishev: Remove args until there is a need for this

sdk-setup/src/mb2
```diff
@@ -1675,6 +1676,22 @@ switch_to_snapshot() {
     notice "Using the '$OPT_TARGET' snapshot of the build target"
 }
 
+sign_rpms() {
+    [[ -e $STATEDIR/spec ]] || fatal "No previoud build found"
```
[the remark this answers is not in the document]
etaishev: Fixed

sdk-setup/src/mb2
```diff
@@ -1675,6 +1676,22 @@ switch_to_snapshot() {
...
+    rpms=$(find_rpms $all_opt "$OPT_OUTPUTDIR" "$STATEDIR/spec")
```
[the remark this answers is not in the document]
etaishev: Added

etaishev:
Add functions for generatind and deploy gpg keys

sdk-setup/src/mb2
```diff
@@ -309,6 +309,24 @@ COMMANDS
...
+
+         -i|--install
+            If install option enabled, key will be exported to rpm and passphrase exported to .....
+
+    key-list
+        Show GPG keys, installed in rpm
+
+    deploy-key -n|--name[=<GPG key name>]
+        Deploy all GPG key to a device with name <name>.
+
```
martyone, line 329: The user should be able to fully control all aspects of their keys creation and so I am certain mb2 should not provide commands for key management and only allow to use existing keys - created with the tool of user's choice or supplied at organization level, etc.

mb2 may be running under the build engine as well, while the keys would be stored on host, so gpg needs to be pointed to the correct location under build engine (`--homedir "$HOME/share/.gnupg"`?).

The key should be then passed with the help of a global option instead of a subcommand. We can think about the proper naming later (it needs to play nicely with sfdk whose domain is much wider and so naming really matters there), and let's work with this meanwhile: `--signing-key <ID-or-file> --signing-passphrase[-file] <passphrase>`.

The `deploy-key` command shouldn't be added either. Similarly to key generation, key deployment is rarely done and well supported with existing tools. So it is enough to add a hint to the documentation of the `--signing-key` option, that the key may be deployed to a device with command like:

`gpg2 --export -a <ID> |mb2 run sudo rpmkeys --import -`

And equivalently with sfdk:

`gpg2 --export -a <ID> |device exec sudo rpmkeys --import -`


For that we need to fix `rpmkeys` to allow reading the key from stdin (see https://github.com/rpm-software-management/rpm/blob/bfee1410af51c1cc9724791fb8d985260a62102b/lib/rpmchecksig.c#L85).

sdk-setup/src/mb2
```diff
@@ -1676,6 +1682,35 @@ switch_to_snapshot() {
     notice "Using the '$OPT_TARGET' snapshot of the build target"
 }
 
+run_key_list() {
+    rpm -q gpg-pubkey --qf '%{name}-%{version}-%{release} --> %{summary}\n'
```
martyone, line 1686: I believe this was meant to list keys on the device side.

martyone:
GPG version on host and under the SDK (chroot or build engine) are likely incompatible. This may have two consequences:

1. GPG under SDK may do unwanted conversion of the data, rendering them useless on host
2. GPG under SDK may not be able to work with the data managed by GPG on host

So we cannot really access single GPG home from both the host and the SDK. We must keep separate storages on host and the SDK, the primary storage being on host, and synchronize the keys to the SDK as needed.

Under SDK chroot we should set `GNUPGHOME=$HOME/.gnupg.sdk` in the environment and also `mount -o bind,ro $HOME/.gnupg $HOME/.gnupg` to be really sure we do not tamper with the GPG data on host.

SDK chroot users can synchronize keys with native tools. I do not think we should provide any supporting tools for them.

`gpg --export-secret-keys ID |sdk gpg2 --import signing.key
`

In contrast to SDK chroot, application SDK should transfer the key automatically for the user - there should be as little as possible notion of build engine existence. For the same reason the user should be also able to select the signing key without the need to edit `~/.rpmmacros` (under the build engine). As a UI to these two functions we should add a new configuration option `package.signing-user <name>` to sfdk and corresponding `--package-signing-user <name>` global option to mb2.

The rationale for naming it `package.signing-user` instead of `package.signing-key` is that it should correspond to the `--local-user` option of gpg2 which is utilized by rpmsign in the default configuration.

When the `package.signing-user <name>` option is set, sfdk should ensure that the corresponding private key is exported as `~/SailfishSDK/vmshare/gnupg/<name>.key` prior to executing a command on the build engine.

mb2, when used with the `--package-signing-user <name>` option, should pass `--define "_gpg_name <name>"` to rpmsign. Additionally, when used under the build engine, it should import the key from `/etc/mersdk/share/gnupg/<name>.key` unless done before.

etaishev:
> GPG version on host and under the SDK (chroot or build engine) are likely incompatible. This may have two consequences:
> 
> 1. GPG under SDK may do unwanted conversion of the data, rendering them useless on host
> 2. GPG under SDK may not be able to work with the data managed by GPG on host
> 
> So we cannot really access single GPG home from both the host and the SDK. We must keep separate storages on host and the SDK, the primary storage being on host, and synchronize the keys to the SDK as needed.
> 
> Under SDK chroot we should set `GNUPGHOME=$HOME/.gnupg.sdk` in the environment and also `mount -o bind,ro $HOME/.gnupg $HOME/.gnupg` to be really sure we do not tamper with the GPG data on host.
> 
> SDK chroot users can synchronize keys with native tools. I do not think we should provide any supporting tools for them.
> 
> `gpg --export-secret-keys ID |sdk gpg2 --import signing.key `
> 
> In contrast to SDK chroot, application SDK should transfer the key automatically for the user - there should be as little as possible notion of build engine existence. For the same reason the user should be also able to select the signing key without the need to edit `~/.rpmmacros` (under the build engine). As a UI to these two functions we should add a new configuration option `package.signing-user <name>` to sfdk and corresponding `--package-signing-user <name>` global option to mb2.
> 
> The rationale for naming it `package.signing-user` instead of `package.signing-key` is that it should correspond to the `--local-user` option of gpg2 which is utilized by rpmsign in the default configuration.
> 
> When the `package.signing-user <name>` option is set, sfdk should ensure that the corresponding private key is exported as `~/SailfishSDK/vmshare/gnupg/<name>.key` prior to executing a command on the build engine.
> 
> mb2, when used with the `--package-signing-user <name>` option, should pass `--define "_gpg_name <name>"` to rpmsign. Additionally, when used under the build engine, it should import the key from `/etc/mersdk/share/gnupg/<name>.key` unless done before.

For the Application SDK I thought there would be the following scenario:
During build BuildEngine (may be setup-buildengine step?) a key is generated and installed in the rpm. Also add key to ~/.rpmmacros
When need sign with custom key use `--package-signing-user <name> `option, yes

This key is exported to the emulator during installation of it
So the user doesn't need to do anything to deploy applications without confirmation during the first installation

martyone:
I would rather see the confirmation dialog simply disabled on emulator.

etaishev:
Why we need install gpg home? Windows machines will have problems using the shared folder for it

martyone:
> Why we need install gpg home? Windows machines will have problems using the shared folder for it.

I didn't suggest to access GPG home through a shared folder. I even clearly stated this is not an option (no matter what the platform is).

> I would also like to raise the issue of key management.
> Why does the user need so many management methods? If we need the rpm signature for deployment without a dialog, it would be better to inform the user about it as little as possible, leaving only service commands for installing the key on the device. Use only the key that was originally generated in build engine

Even if we need the RPM signature *just* for deployment without a dialog... any key installed on the device denotes an identity of a trusted software publisher, someone whose software may be installed without notice. That's not a use case for a key generated and installed on demand without user notice. Even if the user is content with default cryptographic properties, they should be able to 1) let the key reflect their real identity, 2) manage key lifetime and 3) secure the access to the key at least.

This does not mean we cannot provide a smooth user experience with as little disturbance as possible - we will allow deployment of an existing or a new GPG key during device registration in QtC much like we are doing with SSH keys already. Developers content with the default cryptographic properties will just fill in their name, email, description and let it create a new key pair on their regular keyring. If a key was deployed to a device before, it will be suggested as the default option. And as I mentioned before, on emulators the confirmation dialog should be simply disabled.

etaishev:
Update PR. Add --package-signing-user and --package-signing-passphrase.
Do not add default import key by name from --package-signing-user, because file name and real key id may be different

sdk-setup/src/mb2
```diff
@@ -476,6 +476,16 @@ COMMANDS
...
+    sign {--set-key | --deploy-key | --import-key |--list-keys }
+        --set-key -n|--name[=<GPG key name>] -p|--passphrase[=<key passphrase>]
+            Set default key for signing rpm packages.
+        --import-key -k|--key[=<GPG key file>] -p|--passphrase[=<key passphrase>]
+            Import local key file to use signing rpm packages.
+        --deploy-key -n|--name[=<GPG key name>]
+            Deploy <name> GPG pubkey to <device>.
+        --list-keys
+            Show all known GPG keys. If device is specified, shows its keys.
+
```
martyone, line 488: Covered before. There is no need for such command(s) on mb2 side. Let me repeat:

`--set-key`

The same is achieved with `alias mb2="mb2 --package-signing-user ... --package-signing-passphrase[-file] ..."` under platform SDK and with `sfdk config --global package.signing-user ...` with Application SDK.

`--import-key`

Platform SDK users may import their keys with `gpg2 --export-secret-keys ID |sdk gpg2 --import` - no need to assist that with yet another command. For Application SDK, when the `package.signing-user <name>` option is set, sfdk should ensure that the corresponding private key is exported as `~/SailfishOS/vmshare/gnupg/<name>.key` prior to executing a command on the build engine, where it should be taken into use by mb2.

`--deploy-key`
`--list-key`

Deploying and listing keys is straightforward already, no need to hide it behind mb2/sfdk:

```
gpg2 --export ID |ssh device sdk-deploy-key -
ssh device sdk-deploy-key --list
```

Similarly with sfdk's assistance for Application SDK users

```
gpg2 --export ID |sfdk device exec sdk-deploy-key -
````

People should be allowed to manage keys on their devices even without the need to use an SDK - and `sdk-deploy-key` already allows that. WRT that it might be also better to name it differently, so that it does not indicate (falsely) an exclusive relation to SDK. I would consider e.g. `rpm-deploy-key`.

File: sdk-setup/src/mb2 [hunk 28]
```diff
@@ -1799,7 +1799,25 @@ sign_rpms() {
...
+            if [ $? -ne 0 ]; then
+                notice "GPG key has passphrase protection. Set correct passphrase with \"--package-signing-passphrase\" option."
+                exit $?
+            fi
+
+            echo -e "${OPT_PACKAGE_SIGN_PASSPHRASE}\n${OPT_PACKAGE_SIGN_PASSPHRASE}\n" | rpm --addsign ${rpms} --define "_gpg_name $OPT_PACKAGE_SIGN_USER" \
+                --define "__gpg_sign_cmd %{__gpg} gpg --no-verbose --batch --passphrase-fd 0 -u '%{_gpg_name}' -sbo %{__signature_filename} %{__plaintext_filename}"
+        else
+            rpm --addsign ${rpms}
+        fi
```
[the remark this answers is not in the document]
martyone: I forgot to ask: Is it *really* needed to pass passphrase 2 times? (can be done with `3<<<"$passphrase"$'\n'"$passphrase"` if so).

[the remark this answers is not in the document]
etaishev: `--passphrase-fd 3" 3<<<"$passphrase"`

Dosn't work with rpmsign --define and other ways to use --passphrase-fd 3 I couldn't run on the current version of rpm and gpg

Two times password need because rpmsign ask password twice per file

sdk-setup/src/mb2
```diff
@@ -1799,7 +1799,34 @@ sign_rpms() {
...
+            fi
+
+            echo -e "${passphrase}\n${passphrase}\n" | rpm --addsign ${rpms} --define "_gpg_name $OPT_PACKAGE_SIGN_USER" \
+                --define "_gpg_sign_cmd_extra_args --batch --passphrase-fd 0"
+        else
+            rpm --addsign ${rpms}
+        fi
     fi
     local rc=$?
     [[ $rc -eq 0 ]] || { notice "Failed to sign packages"; return $rc; }
```
martyone: > Two times password need because rpmsign ask password twice per file

And we are not limited to packages that produce single RPM, so we'd need the passphrase repeated N*2 times... no way - we need a passphrase-file instead.

Was there any reason why use used `rpm` instead of `rpmsign`? If no, I would prefer using that.

```suggestion
        sign_rpms_native ${rpms}
    fi
    local rc=$?
    [[ $rc -eq 0 ]] || { notice "Failed to sign packages"; return $rc; }
}

sign_rpms_native() {
    local rpms=("$@")

    local tmpdir=
    sign_rpms_native_cleanup() (
        trap 'echo cleaning up...' INT TERM HUP
        if [[ $tmpdir ]]; then
            rm -rf "$tmpdir"
        fi
    )
    trap 'sign_rpms_native_cleanup; trap - RETURN' RETURN
    trap 'return 1' INT TERM HUP

    local extra_args=()

    if [[ $OPT_PACKAGE_SIGN_USER ]]; then
        if ! gpg2 --list-secret-keys "$OPT_PACKAGE_SIGN_USER" &>/dev/null; then
            notice "Cannot sign packages: No GPG key found for the user '$OPT_PACKAGE_SIGN_USER'."
            return 1
        fi

        extra_args+=(--define "_gpg_name $OPT_PACKAGE_SIGN_USER")

        local passphrase=
        if [[ $OPT_PACKAGE_SIGN_PASSPHRASE_FILE ]]; then
            passphrase=$(<"$OPT_PACKAGE_SIGN_PASSPHRASE_FILE")
        else
            passphrase=$OPT_PACKAGE_SIGN_PASSPHRASE
        fi

        if ! gpg2 --batch --status-fd 1 --sign --local-user "$OPT_PACKAGE_SIGN_USER" \
                --output - --passphrase-fd 3 3<<<"$passphrase" </dev/null &>/dev/null; then
            if [[ ! $passphrase ]]; then
                notice "Cannot sign packages: The selected GPG key is passphrase protected" \
                    "and no passphrase was specified."
            else
                notice "Cannot sign packages: The given passphrase is not a correct passphrase" \
                    "for the '$OPT_PACKAGE_SIGN_USER' key."
            fi
            return 1
        fi

        if [[ $passphrase ]]; then
            tmpdir=$(mktemp -d) || return
            chmod 0700 "$tmpdir" || return

            local passphrase_file=$tmpdir/passphrase
            cat >"$passphrase_file" <<<"$passphrase" || return

            extra_args+=(--define "_gpg_sign_cmd_extra_args --batch --passphrase-file '$passphrase_file'")
        fi
    fi

    rpmsign --addsign "${rpms[@]}" "${extra_args[@]}" </dev/null
```

martyone (commented):
Please ensure that all white space is preserved when applying suggestion from GitHub review. For some reason all the blank lines I used in my suggestions were gone whenever you applied them.

sdk-setup/src/mb2
```diff
@@ -1791,20 +1791,88 @@ run_build_requires__process_args() {
...
+            if [[ -f $key_path ]]; then
+                if ! gpg2 --import "${key_path}" &> /dev/null; then 
+                    notice "Cannot sign packages: Failed to import key from file '$key_path'."
+                    return 1
+                fi
+            else
+                notice "Cannot sign packages: No GPG key found for the user '$OPT_PACKAGE_SIGN_USER' and no GPG key file '$key_path'."
+                return 1
+            fi
+        fi
```
martyone, line 1840: We should be specific about the import functionality being limited to build engine and we also should try harder - what if the user creates a new key pair using the same name? We also cannot simply import the new key without removing the old one first - we would end up with multiple keys matching the user name under the build engine and how would rpmsign tell which one to use?

```suggestion
        if inside_build_engine; then
            local shared_key=${OPT_SHARED_DIR}/gnupg/${OPT_PACKAGE_SIGN_USER}.key
            if [[ ! -f $shared_key ]]; then
                notice "Cannot sign packages: Internal error: No GPG key shared for the user '$OPT_PACKAGE_SIGN_USER'."
                return 1
            fi

            ... import new / update existing ...
        fi
        
        if ! gpg2 --list-secret-keys "$OPT_PACKAGE_SIGN_USER" &>/dev/null; then
            notice "Cannot sign packages: No GPG key found for the user '$OPT_PACKAGE_SIGN_USER'."
            return 1
        fi
```

etaishev: And this is the reason why I didn't want to support import in the signature function. It's not obvious. Why, if there are no key files, we should log out and not sign what is already in the system with this name?

martyone: Application SDK users use QtC (sfdk) and the build engine is a black box to them. They have told QtC (sfdk) which key should be used for signing packages and that's all they need to care about. It is QtC's (sfdk's) responsibility to pass the keys to mb2 and if it fails to, it is an internal error and mb2 has no option to recovery from that.

sdk-setup/src/mb2
```diff
@@ -1791,20 +1791,103 @@ run_build_requires__process_args() {
...
+                if ! gpg2 --batch --delete-secret-keys "$fingerprint"; then
+                    notice "Cannot sign packages: Failed to delete existing '$fingerprint' key."
+                    return 1
+                fi
+            done
+
+            if ! gpg2 --import "${shared_key}" &> /dev/null; then 
+                notice "Cannot sign packages: Failed to import key from file '$shared_key'."
+                return 1
+            fi
```
martyone, line 1849: - local variables
- error handling
- gpg2 use without --with-colons - output format not stable
- wrong dereference of array variable
- public keys not deleted
- use gpg2 --import with --quiet instead of suppressing all output, including possibly valuable error output

```suggestion
            local shared_key_id=$(gpg2 --with-colons --list-packets "$shared_key" \
                    |sed -n 's/:user ID packet: "\(.*\)"/\1/p')
            [[ $shared_key_id ]] || return

            local key_fingerprints=$(gpg2 --with-colons --list-secret-keys --fingerprint "$shared_key_id" \
                    |awk -F: '$1 == "fpr" { print $10 }')
            local key_fingerprint=
            for key_fingerprint in $key_fingerprints; do
                if ! gpg2 --batch --delete-secret-and-public-key "$key_fingerprint"; then
                    notice "Cannot sign packages: Internal error: Failed to delete old GPG key copy '$key_fingerprint'."
                    return 1
                fi
            done

            if ! gpg2 --quiet --import "${shared_key}"; then 
                notice "Cannot sign packages: Internal error: Failed to import GPG key from file '$shared_key'."
                return 1
            fi
```

martyone (commented):
We still miss setting GNUPGHOME="$HOME/.gnupg.sailfishsdk" somewhere system-wide.

martyone:
Documentation to add after the MAINTAINING CHANGELOGS section (two blank lines before, one after): 

```

SIGNING PACKAGES

The 'build' and 'package' commands allow to optionally sign the resulting
packages. Use the '--sign' option to these commands to enable this step.

The cryptographic key to use for signing can be selected with the global
option '--package-signing-user'. If this is a passphrase-protected key, the
passphrase may be supplied with one of the '--package-signing-passphrase' or
'--package-signing-passphrase-file' options. Set up yourself a shell alias to
apply these by default:

    $ touch PASSFILE
    $ chmod 600 PASSFILE
    $ cat >PASSFILE
    YourPassphrase
    ^D
    $ alias mb2="mb2 --package-signing-user NAME \
        --package-signing-passphrase-file PASSFILE"

The GPG version on host and under the SDK are not likely to match - for that
reason the SDK uses its own GPG home directory and therefore your secret key
must be imported under the SDK before it can be used. This can be achieved
with command like this (executed on host):

    $ gpg --export-secret-keys ID |sdk gpg2 --import

(Replace 'sdk' with the command you use to enter the SDK)

```

Documentation to add at the very end of the global options documentation:

```
    --package-signing-user <name>
        Select the GPG key to sign packages with. Accepts the same values
        as the '--local-user' option of the gpg command does. See GPG
        documentation for more details.

        Depending on your SDK variant, different semantics may apply.

    --package-signing-passphrase <string>
        In case a passphrase-protected signing key is selected for signing
        packages, use <string> as the passphrase.
        
    --package-signing-passphrase-file <file>
        In case a passphrase-protected signing key is selected for signing
        packages, read the passphrase from the <file>. Only the first line will
        be read.
```

etaishev:
> We still miss setting GNUPGHOME="$HOME/.gnupg.sailfishsdk" somewhere system-wide.

Should we take care of the existence of this folder or leave it to the user's discretion as well?

sdk-setup/src/mer-sdk-chroot
```diff
@@ -291,6 +293,11 @@ prepare_user() {
     fi
     echo "$user ALL=NOPASSWD: ALL" > ${sdkroot}/etc/sudoers.d/$user
     chmod 0440 ${sdkroot}/etc/sudoers.d/$user
+
+    # mount ~/.gnupg (default GNUPGHOME) as read only
+    if [[ -d "${sdkroot}${HOMEDIR}/.gnupg" ]]; then
+        mount --bind -o ro $HOMEDIR/.gnupg ${sdkroot}${HOMEDIR}/.gnupg
+    fi
```
martyone, line 300: I would be more specific about our motivation to do this, so let me suggest a small adjustment (moving under the preceding `if` and bind mounting the directory onto itself):

```suggestion

        # Prevent accidental writes to host's GPG home (SDK uses its own GPG home)
        if [[ -d ${sdkroot}${HOMEDIR}/.gnupg ]]; then
            mount --bind -o ro "${sdkroot}${HOMEDIR}/.gnupg"{,}
        fi
    fi

    echo "$user ALL=NOPASSWD: ALL" > ${sdkroot}/etc/sudoers.d/$user
    chmod 0440 ${sdkroot}/etc/sudoers.d/$user
```

sdk-setup/src/mer-sdk-chroot
```diff
@@ -396,6 +403,9 @@ setarch i386 chroot ${sdkroot} /bin/su -s /bin/bash -l $user -- -c "$(cat <<END
 export SSH_AUTH_SOCK='${MY_SSH_AUTH_SOCK}'
 export SSH_AGENT_PID=${SSH_AGENT_PID:-}
 export MERSDK='$sdkroot'
+export GNUPGHOME='${HOMEDIR}/${SFDK_GNUPGHOME_DIR}'
+
+[[ ! -d "${HOMEDIR}/${SFDK_GNUPGHOME_DIR}" ]] && mkdir -p "${HOMEDIR}/${SFDK_GNUPGHOME_DIR}" && sudo chmod 700 ${HOMEDIR}/${SFDK_GNUPGHOME_DIR}
```
martyone, line 408: I would create it only when we need it (from mb2), so that we do not pollute user's home unnecessarily.

This answers your question "Should we take care of the existence of this folder or leave it to the user's discretion as well?" I believe.

sdk-setup/src/mer-sdk-chroot
```diff
@@ -75,6 +75,8 @@ if cmp -s /proc/$PPID/mountinfo /proc/self/mountinfo; then
     exit 1
 fi
 
+SFDK_GNUPGHOME_DIR='.gnupg.sailfishsdk'
+
```
martyone, line 79: It should be SDK_, not SFDK_ in this context, and it would be better moved closer to the top, maybe just before MY_SSH_AUTH_SOCK and maybe name it similarly - MY_GNUPGHOME?

sdk-setup/src/mer-sdk-chroot
etaishev: I agree, but if you just run `gpg2 --import` (any gpg2 command), you will get an error saying that there is no `GNUPGHOME`. May be create `GNUPGHOME` under `/srv/mer/sdks/sfossdk/`? Then it will be deleted every time reinstall it

martyone: yeah, maybe better use `$HOME/.config/SailfishSDK/gnupg`

sdk-setup/src/mer-sdk-chroot
```diff
@@ -283,12 +284,18 @@ prepare_user() {
     install --owner $user --group $(id -gn $user) -d ${sdkroot}${HOMEDIR}
 
     if [[ $bind_mount_home == "yes" ]] ; then
-	echo "Mounting home directory: ${HOMEDIR}"
-	mount --bind ${HOMEDIR} ${sdkroot}${HOMEDIR}
-	# Now the sdk uses a private namespace, there's no need to
-	# make it unbindable
-	mount --make-shared ${sdkroot}${HOMEDIR}
+    	echo "Mounting home directory: ${HOMEDIR}"
+        mount --bind ${HOMEDIR} ${sdkroot}${HOMEDIR}
```
martyone, line 288: Bad indent, avoid whitespace change (keep tabs in old code, use spaces in new code).

sdk-setup/src/mer-sdk-chroot
```diff
@@ -396,6 +403,9 @@ setarch i386 chroot ${sdkroot} /bin/su -s /bin/bash -l $user -- -c "$(cat <<END
 export SSH_AUTH_SOCK='${MY_SSH_AUTH_SOCK}'
 export SSH_AGENT_PID=${SSH_AGENT_PID:-}
 export MERSDK='$sdkroot'
+export GNUPGHOME='${HOMEDIR}/${MY_GNUPGHOME_DIR}'
+
+[[ ! -d "${HOMEDIR}/${MY_GNUPGHOME_DIR}" ]] && mkdir -p "${HOMEDIR}/${MY_GNUPGHOME_DIR}" && sudo chmod 700 ${HOMEDIR}/${MY_GNUPGHOME_DIR}
```
martyone, line 408:
```suggestion
install --directory --mode 700 "\$GNUPGHOME"
```

martyone (commented):
Please squash "[mb2] Add package-signing-user option. JB#50188" into "[mb2] Allow to sign created packages. JB#50188".

martyone:
I would prefer the following summary for the last commit:

[mer-sdk-chroot] Prevent accidental writes to host's GPG home

sdk-setup/src/mer-sdk-chroot
```diff
@@ -396,6 +402,9 @@ setarch i386 chroot ${sdkroot} /bin/su -s /bin/bash -l $user -- -c "$(cat <<END
 export SSH_AUTH_SOCK='${MY_SSH_AUTH_SOCK}'
 export SSH_AGENT_PID=${SSH_AGENT_PID:-}
 export MERSDK='$sdkroot'
+export GNUPGHOME='${HOMEDIR}/${MY_GNUPGHOME_DIR}'
+
+install --directory --mode 700 "\$GNUPGHOME"
```
martyone, line 407: Last nitpick - sorry I didn't realize this earlier - this would be better done from prepare_user(), just after the `if [[ $bind_mount_home == "yes" ]] ; then ... fi` block:

```
install --owner $user --group $(id -gn $user) --mode 700 -d ${sdkroot}${HOMEDIR}/${MY_GNUPGHOME_DIR}
```

(omitted quotes as the surrounding code does :\)

martyone (commented):
Last touch and this LGTM :)

Now the qtc side needs to be updated at least for the "Do not imply package creation" commit before this can be merged.

martyone (approved):
QtC side needs to be updated at least for the "Do not imply package creation" commit before this can be merged.

etaishev:
Update creating `GNUPGHOME`. `install -d` apply owner and mode only for installed dir: `gnupg`, but `SailfishSDK` stay with root owner ( if it was not created earlier )

martyone (approved):
QtC side needs to be updated at least for the "Do not imply package creation" commit before this can be merged.

sdk-setup/src/mb2
```diff
@@ -1809,6 +1855,106 @@ run_build_requires__process_args() {
     fi
 }
 
+sign_rpms() {
+    [[ -e $STATEDIR/spec ]] || fatal "No previous build found"
+    local rpms=$(find_rpms $all_opt "$OPT_OUTPUTDIR" "$STATEDIR/spec")
+    [[ $rpms ]] || fatal "No package found"
+
+    local sdksign=/usr/libexec/sdk-setup/sdk-rpmsig
```
martyone, line 1863: Typo

```suggestion
    local sdksign=/usr/libexec/sdk-setup/sdk-rpmsign
```

martyone:
It seems during last rebase you lost the "[mb2] Do not imply package creation. JB#50188" commit https://github.com/sailfishos/sdk-setup/commit/075362c246af2314064193b95800316e53813126 . The calls to run_make_install and run_package are still there, they were just moved around, so the commit is still needed.

sdk-setup/src/mb2
```diff
@@ -200,6 +200,35 @@ section in the spec file.  If a file with '.changes.run' extension is
...
+passphrase may be supplied with one of the '--package-signing-passphrase' or
+'--package-signing-passphrase-file' options. Set up yourself a shell alias to
+apply these by default:
+
+    $ touch PASSFILE
+    $ chmod 600 PASSFILE
+    $ cat >PASSFILE
+    YourPassphrase
+    ^D
+    $ alias mb2="mb2 --package-signing-user NAME \
```
martyone, line 220: Double backslash is needed

```suggestion
    $ alias mb2="mb2 --package-signing-user NAME \\
```

PR #185, opened 2020-07-14
[mb2] Remove trailing CRs from patch ops and filenames. JB#49598

vigejolla:


vigejolla:
> I would prefer aligning with the other command implementations (e.g. run_build), using a tmp copy of the spec and fixing line ends in whole file.

Understandable. Changed to use that approach.

vigejolla:
> I would still consider making it use the _very_ same approach as the others do - copy first, then modify with sed. Easy to spot (visual) patterns are invaluable :)

```
run_build() {
    # intended to provide mb build behaviour
    maybe_verify_target_dependencies

    maybe_increment_build_number;
    if [[ $BUILD_NOPREP ]]; then
        sed -e '/^%patch/d' "$OPT_SPEC" > "$TMP_SPEC"
```

martyone:
yeah, run_build is not the best example but the others do well

PR #186, opened 2020-07-14
[sdk-assistant] Check target/tooling existence. Fixes JB#48753 

vigejolla:


sdk-setup/src/sdk-assistant
```diff
@@ -383,6 +383,9 @@ if [[ $OPT_PACKAGE_LIST ]]; then
 fi
 
 if [[ $OPT_UPDATE ]]; then
+    if ! have "$OPT_TYPE" "$OPT_NAME"; then
+        fail "$OPT_NAME: $OPT_TYPE does not exist"
+    fi
```
martyone, line 388: Shouldn't we check it for other commands too? If so, I would move it as an `elif` branch to the `if` block closing at line 375. I would prefer saying `"$OPT_NAME: No such $OPT_TYPE"`.

vigejolla: Makes sense - changed so

PR #187, opened 2020-07-15
[sdk-assistant] Add --tooling parameter. Fixes JB#49105

vigejolla:


sdk-setup/src/sdk-assistant
```diff
@@ -51,9 +51,11 @@ usage() {
 SDK assistant tool
 
 Usage:
-    $(basename $0) [tooling|target] create <name> <URL>
+    $(basename $0) [tooling|target] create <name> <URL> [--tooling <tooling>]
                             : create a new sb2 tooling or target <name>,
-                              using the image at <URL>.
+                              using the image at <URL>. It is possible to 
+                              force tooling selection when creating a target
+                              with '--tooling'.
```
martyone, line 58: And here is the suggestion: "In case of a new target, an appropriate tooling will be selected automatically unless overridden with '--tooling'." :)

PR #191, opened 2020-08-19
Jb50814

martyone:


martyone:
It is still to be determined if the error affects the Docker based build engine.

Just the second commit belongs to this PR.

vigejolla (approved):
This most likely has negative impact on performance. But I'll take slow performance over read errors every time.

martyone:
> This most likely has negative impact on performance. But I'll take slow performance over read errors every time.

Fortunately the performance is about the same, maybe negligibly better :)

PR #192, opened 2020-08-20
[mb2] Ignore default ssh keys which require password. Fixes JB#50884

vigejolla:
If any of the default keys requires password, ssh-add would fail. Now
we print a warning instead.

### DIFF
--- a/sdk-setup/src/mb2
+++ b/sdk-setup/src/mb2
@@ -1120,6 +1120,11 @@ with_specific_or_default_keys() {
     fi
 
     add_specific_or_default_keys() {
+
+        ssh-add() {
+            DISPLAY=:0 SSH_ASKPASS=/usr/libexec/sdk-setup/ssh-askpass command ssh-add "$@"
+        } 
+
         local key=$1
         if [[ $key ]]; then
             ssh-add -q - <"$key" || return
@@ -1130,7 +1135,7 @@ with_specific_or_default_keys() {
             # genericlinuxdeviceconfigurationwizardpages.cpp
             for default_key in "$real_home"/.ssh/id_{rsa,ecdsa,ed25519}; do
                 if [[ -f $default_key ]]; then
-                    ssh-add -q - <"$default_key" # allow failure
+                    ssh-add -q - <"$default_key" || :
                 fi
             done
         fi
--- a/rpm/sdk-setup.spec
+++ b/rpm/sdk-setup.spec
@@ -203,6 +203,7 @@ cp src/sdk-init %{buildroot}%{_bindir}/
 mkdir -p %{buildroot}%{_libexecdir}/%{name}
 cp src/oomadvice %{buildroot}%{_libexecdir}/%{name}/
 cp src/sdk-setup-swap %{buildroot}%{_libexecdir}/%{name}/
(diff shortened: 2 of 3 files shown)


sdk-setup/src/mb2
```diff
@@ -1130,7 +1130,8 @@ with_specific_or_default_keys() {
             # genericlinuxdeviceconfigurationwizardpages.cpp
             for default_key in "$real_home"/.ssh/id_{rsa,ecdsa,ed25519}; do
                 if [[ -f $default_key ]]; then
-                    ssh-add -q - <"$default_key" # allow failure
+                    ssh-add -q - <"$default_key" \
+                        || echo "Warning: using keys with passwords is not possible"
```
martyone, line 1134: It may fail for other reasons as well, so I would try to isolate the specific case here.

One possibility would be to provide a dummy `/usr/libexec/sdk-setup/ssh-askpass` that would only write a message and exit with nonzero, then create a wrapper for `ssh-add` under `add_specific_or_default_keys`.

```
ssh-add() {
    DISPLAY=:0 SSH_ASKPASS=/usr/libexec/sdk-setup/ssh-askpass command ssh-add "$@"
}
```

IIUC the error would be fatal in both cases (specific or default key), so we can raise the warning to error with `/usr/libexec/sdk-setup/ssh-askpass`:

```
#!/bin/bash
echo "Fatal: Use of password-protected SSH keys is not supported" >&2
exit 1
```

vigejolla: Sure, we can go that route as well. I'm going to "downgrade" the error message though - it is not fatal to have password protected default keys, as long as there is one working default key without password.

martyone: > ... it is not fatal to have password protected default keys, as long as there is one working default key without password.

hm, under bz I read

> If connection to device is configured to use a default key (id_rsa, id_ecdsa or id_ed25519), and any of the default keys requires password, deployment to device fails. This happens even if the actually used key does not use password.

martyone:
Commit message/PR summary need update.

sdk-setup/src/mb2
vigejolla: Yes, that is a bug and this PR is about fixing that bug - making the deployment succeed even if there is a default key with password.

vigejolla:
> Commit message/PR summary need update.

To me it seems it's still quite correct

sdk-setup/src/mb2
martyone: So it really fixes this? I though it was only about issuing the warning and I still don't see how it fixes that - explain please :)

vigejolla: See that "return 0" in the end of ssh-add?

martyone: Ah that way :)

I would prefer leaving the `return 0` out and do this instead:

`ssh-add -q - <"$default_key" || :`

vigejolla: > `ssh-add -q - <"$default_key" || :`

I thought about doing that way but decided to go with return 0 instead. I can't remember why though - perhaps '||  :' just looked silly to me.

martyone:
> > Commit message/PR summary need update.
> 
> To me it seems it's still quite correct

Indeed

martyone (approved):

PR #193, opened 2020-08-26
[mb2] Use binary mode when applying patches. Fixes JB#49598

vigejolla:
By default, when patch notices CRLF line endings in patches, it removes
the CRs. As the files to be patched still have the CRLF line endings,
patch fails. Using binary mode disables the heuristics.

### DIFF
--- a/sdk-setup/src/mb2
+++ b/sdk-setup/src/mb2
@@ -1713,6 +1713,7 @@ run_apply() {
     fi
 
     tr -d '\r' < "$OPT_SPEC" > "$TMP_SPEC"
+    cmp -s "$OPT_SPEC" "TMP_SPEC" || common_op+=(--binary)
 
     local to_apply=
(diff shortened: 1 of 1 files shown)


sdk-setup/src/mb2
```diff
@@ -1737,7 +1737,7 @@ run_apply() {
             fi
             patch_file="$(sed -n "s/^Patch$p: \+\(.*\)/\1/p" "$TMP_SPEC")"
             echo "$message $patch_file" >&2
-            patch "${common_op[@]}" $op < "$(dirname "$OPT_SPEC")/$patch_file" >&2
+            patch --binary "${common_op[@]}" $op < "$(dirname "$OPT_SPEC")/$patch_file" >&2
```
martyone, line 1740: The option should be added to `common_op`. I would be more assured if we added it conditionally, e.g., only if `$OPT_SPEC` contains CRLF line endings.

martyone (approved):
What an elegant fix!

PR #194, opened 2020-08-27
Synchronise /usr/lib64 of 64 bit targets

vigejolla:
On 64 bit targets, all files which need to be synchronised are stored in /usr/lib64 instead of /usr/lib

### DIFF
--- a/sdk-setup/src/sdk-manage
+++ b/sdk-setup/src/sdk-manage
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# This is a helper script for the Mer SDK to manage sb2 target and
+# This is a helper script for the Sailfish SDK to manage sb2 target and
 # toolchain installation
 
 set -o nounset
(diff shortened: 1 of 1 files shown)


sdk-setup/src/sdk-manage
```diff
@@ -1310,6 +1310,10 @@ list_targets() {
 
 synchronise_target() {
     local target=$1
+    local libdir=/usr/lib
+    if [[ $target == *64 ]]; then
```
martyone, line 1314:
```suggestion
    if [[ -d $MER_TARGETS/$target/usr/lib64 ]]; then
```

I would also move it below the `target_is_snapshot` check.

martyone (commented):
Typo `[sdk-manager]` in second commit. I would omit `[]` from the first commit. I didn't see JB#xxxxx reference IIRC.

martyone:
Wrong branch name: 58055 -> 50855

vigejolla:
Opened new pull request from correct branch

PR #196, opened 2020-09-10
[sdk-manage] Replace rpmdb in two steps on overlay filesystems. JB#49910

vigejolla:
The overlay filesystem does not support overwriting a directory in a
single step, like rpm does after rebuilding the db. Luckily, rpm leaves
the newly built db directory in place, so we can just remove the old
one and rename the new one after that.

### DIFF
--- a/sdk-setup/src/sdk-manage
+++ b/sdk-setup/src/sdk-manage
@@ -611,6 +611,30 @@ object_upgradable() {
     object_zypper "$type:$object" --quiet --no-refresh list-updates
 }
 
+rebuild_rpmdb() {
+    local object=$1
+
+    rebuilddb_output_filter() {
+        output="$(< /dev/stdin)"
+        expected_output='error: failed to replace old database with new database!
+error: replace files in /var/lib/rpm with files from /var/lib/rpmrebuilddb.[0-9]* to recover'
+        [[ "$output" =~ $expected_output ]] || echo "$output"
+    }
+
+    # rpm can't overwrite directory on a different overlay layer
+    if [[ $(findmnt --noheadings --output FSTYPE / ) == overlay ]]; then
+        enter "$object" rpm --rebuilddb 2>&1 | rebuilddb_output_filter
+        local new_db=$(enter "$object" find /var/lib -maxdepth 1 -name 'rpmrebuilddb.*' |head -n1)
+        if [[ $new_db ]]; then
+            # sb2 recreates rpm directory on startup, so we need to do these in
+            # a single session
+            enter "$object" sh -c "rm -rf /var/lib/rpm && mv $new_db /var/lib/rpm"
+        fi
+    else
+        enter "$object" rpm --rebuilddb
+    fi
+}
+
 upgrade_object() {
     local type=$1 object=${2:-}
 
@@ -622,13 +646,13 @@ upgrade_object() {
     # version mismatch" errors e.g. after upgrading RPM
     if [[ $type == target ]]; then
         # Ignore possible errors here in hope that they will disappear after the upgrade
-        enter_target "$object" rpm --rebuilddb
+        rebuild_rpmdb "$type:$object"
     fi
 
     object_zypper "$type:$object" dup || return
 
     if [[ $type == target ]]; then
-        enter_target "$object" rpm --rebuilddb || return
+        rebuild_rpmdb "$type:$object" || return
     fi
 }
 


sdk-setup/src/sdk-manage
```diff
@@ -611,6 +611,21 @@ object_upgradable() {
...
+    if mount | grep -q "overlay on / type overlay"; then
+        enter_target "$object" rpm --rebuilddb > /dev/null 2&>1
+        local new_db=$(enter_target "$object" ls /var/lib|grep rpmrebuild)
+        if [[ ! -z $new_db ]]; then
+            enter_target "$object" rm -Rf /var/lib/rpm
+            enter_target "$object" mv /var/lib/$new_db /var/lib/rpm
+        fi
+    else
+        enter_target "$object" rpm --rebuilddb
+    fi
```
martyone, line 626:
```suggestion
    if [[ $(findmnt --noheadings --output FSTYPE / ) == overlay ]]; then
        enter "$object" rpm --rebuilddb &>/dev/null
        local new_db=$(enter "$object" find /var/lib -maxdepth 1 -name 'rpmrebuilddb.*' |head -n1)
        if [[ $new_db ]]; then
            enter "$object" rm -rf /var/lib/rpm || return
            enter "$object" mv "$new_db" /var/lib/rpm
        fi
    else
        enter "$object" rpm --rebuilddb
    fi
```

sdk-setup/src/sdk-manage
```diff
@@ -622,13 +637,13 @@ upgrade_object() {
     # version mismatch" errors e.g. after upgrading RPM
     if [[ $type == target ]]; then
         # Ignore possible errors here in hope that they will disappear after the upgrade
-        enter_target "$object" rpm --rebuilddb
+        rebuild_rpmdb "$object"
```
martyone, line 640:
```suggestion
        rebuild_rpmdb "$type:$object"
```

sdk-setup/src/sdk-manage
```diff
@@ -622,13 +637,13 @@ upgrade_object() {
...
     fi
 
     object_zypper "$type:$object" dup || return
 
     if [[ $type == target ]]; then
-        enter_target "$object" rpm --rebuilddb || return
+        rebuild_rpmdb "$object" || return
```
martyone, line 646:
```suggestion
        rebuild_rpmdb "$type:$object" || return
```

sdk-setup/src/sdk-manage
vigejolla: Why the change to use enter instead of enter_target?

martyone: Given its name, position in the file and (most importantly) effect, it is a generic function. No matter that we only use it with targets right now.

martyone (approved):

martyone:
One more idea: could we just filter out the expected error messages(*) instead of suppressing all output?

*)
error: failed to replace old database with new database!
error: replace files in /var/lib/rpm with files from
/var/lib/rpmrebuilddb.4682 to recover

vigejolla:
> One more idea: could we just filter out the expected error messages(*) instead of suppressing all output?
> 
> *)
> error: failed to replace old database with new database!
> error: replace files in /var/lib/rpm with files from
> /var/lib/rpmrebuilddb.4682 to recover

Maybe. But we also have a problem - yesterday as I was testing with SFOS 3.4.0.9 (i.e. SDK 3.3.2) this worked. Today, as I'm testing with 3.4.0.17 (i.e. SDK 3.3.3) it no longer works. `rm -rf /var/lib/rpm` does not quite work when run under sb2 - it seems to delete the directory, but as soon as we return from sb2, the directory is back. While testing from command line, executing both rm and mv under a single shell (i.e. `sb2 -t SailfishOS-3.4.0.17-armv7hl -R -m sdk-install sh -c "rm -rf /var/lib/rpm && mv /var/lib/rpmbuilddb.1234 /var/lib/rpm`) seems to do the trick, but for some reason not when it's run as part of sdk-manage...

vigejolla:
Please check if stuff around line 621 feels completely insane to you. I'll think about the filtering a bit later and add it as a separate commit.

sdk-setup/src/sdk-manage
```diff
@@ -611,6 +611,20 @@ object_upgradable() {
...
 }
 
+rebuild_rpmdb() {
+    local object=$1
+    # rpm can't overwrite directory on a different overlay layer
+    if [[ $(findmnt --noheadings --output FSTYPE / ) == overlay ]]; then
+        enter "$object" rpm --rebuilddb &>/dev/null
+        local new_db=$(enter "$object" find /var/lib -maxdepth 1 -name 'rpmrebuilddb.*' |head -n1)
+        if [[ $new_db ]]; then
+            enter "$object" sh -c "rm -rf /var/lib/rpm && mv $new_db /var/lib/rpm"
```
martyone, line 621: Worth a comment that in case of targets we need to do those under single sb2 session and why.

vigejolla:
> > it seems to delete the directory, but as soon as we return from sb2, the directory is back.
> 
> I might have a clue - recreated next time sb2 executes when it evaluates this? https://git.sailfishos.org/mer-core/scratchbox2/blob/master/scratchbox2/modes/obs-rpm-install/fs_rules.lua#L62

Thanks - I was starting to question my own sanity :) It is indeed so - the directory is gone when checking outside sb2, but it is recreated when sb2 executes.

PR #197, opened 2020-09-11
Jb51109

martyone:


sdk-setup/src/mb2
```diff
@@ -174,7 +174,7 @@ testing on one or more of the four recognized test levels:
        the RPM .spec file if present.
 
        Testing at this level is normally performed by the 'build' and
-       'package' commands (unless used with the '--nocheck' option) and so it
+       'package' commands (unless used with the '--no-check' option) and so it
```
vigejolla, line 177: I'm not sure renaming the parameter to --no-check is a good idea. Quite many commands have --no-something parameters which are the opposite of --something. But we don't have a corresponding --check -parameter.

vigejolla (approved):
LGTM, except for different opinion on renaming --nocheck to --no-check, but that is just my opinion

sdk-setup/src/mb2
martyone: Here we just follow the naming convention that --all-words-are-separated no matter how long they are :)

PR #198, opened 2020-09-14
[sdk-manage] Handle zypper exit codes. Fixes JB#51186

Thaodan:


sdk-setup/src/sdk-manage
```diff
@@ -509,6 +509,18 @@ object_zypper() {
     local object=$1 args=("${@:2}")
 
     enter "$object" zypper --non-interactive "${args[@]}"
+    case $? in
+        100|101|102)
+            # Ignored exit codes:
+            # 100 - ZYPPER_EXIT_INF_UPDATE_NEEDED
+            # 101 - ZYPPER_EXIT_INF_SEC_UPDATE_NEEDED
+            # 102 - ZYPPER_EXIT_INF_REBOOT_NEEDED
+            :
```
martyone, line 518: I would consider an explicit `return 0` as a more readable alternative :)

sdk-setup/src/sdk-manage
```diff
@@ -509,6 +509,18 @@ object_zypper() {
...
+            ;;
+        103)
+            # 103 - ZYPPER_EXIT_INF_RESTART_NEEDED
+            object_zypper "$@"
```
martyone, line 522: Worth to care about endless restart loop?

nitpick: I would prefer not omitting `;;` here.

Thaodan: That won't happen unless you push endless zypper updates. After running zypper again the return status is different.

sdk-setup/src/sdk-manage
Thaodan: Can change that however you prefer it. I agree the meaning of : isn't that obvious.

martyone (commented):
Rebase on top of current master please

Thaodan:
done

martyone:
Rebase once again please :)

PR #199, opened 2020-09-17
Jb50666

martyone:


sdk-setup/src/oomadvice
```diff
@@ -111,7 +111,7 @@ subprocess()
 
 set_defaults()
 {
-    OPT_OOM_ADJ=10
+    OPT_OOM_SCORE_ADJ=10
```
vigejolla, line 114: oom_adj range was from -17 to 15, oom_score_adj range is from -1000 to 1000. Should we adjust the default accordingly?

martyone: good catch! I thought they we just aligning the name. Adjusted proportionally.

vigejolla (approved):
LGTM

PR #201, opened 2020-09-24
Jb49899

martyone:


sdk-setup/systemd/sdk-refresh.timer
```diff
@@ -5,7 +5,7 @@
 # The refresh service will ensure it does not run more than once#
 # a day
 OnActiveSec=1d
-OnBootSec=5m
+OnStartupSec=5m
```
vigejolla, line 8: Too bad our systemd is still at v225 - from v228 onwards OnBootSec==OnStartupSec inside containers :/

vigejolla (approved):
LGTM

PR #202, opened 2020-09-24
Jb51082

martyone:


sdk-setup/src/mb2
```diff
@@ -2589,10 +2589,12 @@ run_check__process_args_helper() {
     RPMVALIDATION_ALL_SUITES=($(awk '{print $1}' <<<"$rpmvalidation_suites_info"))
     local rpmvalidation_essential_suites=($(awk -v essential="$CHECK_ESSENTIAL" \
         '($2 == essential) { print $1 }' <<<"$rpmvalidation_suites_info"))
-    CHECK_SUITES=($CHECK_RPMSPEC ${rpmvalidation_essential_suites[*]})
+    local known_suites=($CHECK_RPMSPEC $CHECK_RPMLINT ${RPMVALIDATION_ALL_SUITES[*]})
```
vigejolla, line 2592: Our current targets don't have rpmlint installed. For future releases I think we should add rpmlint to the target images, but in order to support older targets we should do something here. At least we have to check if rpmlint is installed - if it's not we should either install it or not add rpmlint to the list of suites. Or add it to the list but not run it unless explicitly specified.

vigejolla (changes requested):
Otherwise looks good, but we should do something if rpmlint is not installed

sdk-setup/src/mb2
martyone: I have considered those options too and so far I found the current behavior most convenient:

`rpmlint could not be executed: No such file or directory`

If we hid the suite conditionally, discoverability would suffer and "fixing" that with any kind of documentation is something I would like to avoid.

Similarly when we disabled it conditionally - this would require extending the set of possible states.

Trying to get it installed automatically under older targets (toolings) is something where problems are to be expected (network may not be accessible at that time, sometimes repos do not refresh on first try, ...), so I wouldn't even try that.

vigejolla: At least the error message could indicate what the user has to do in order to install rpmlint. That's the **minimum** that we should do.

vigejolla: And I have to say that I disagree with your statement about discoverability - We should add rpmlint to the target images in future releases, so discoverability would no longer be an issue. Assuming your "hid the suite conditionally" means that the suite is hidden if the target does not have rpmlint installed.

martyone: Of course we will add rpmlint to the target (tooling) images in future releases, I was sure we only talk about old releases :)

vigejolla (commented):
This looks quite good, but I don't think it should be merged before we have rpmlint in the images.

vigejolla (approved):
LGTM

PR #203, opened 2020-10-01
[mb2] Warn on CRLF use. JB#51089

martyone:


sdk-setup/src/mb2
```diff
@@ -803,6 +803,24 @@ ensure_spec_newer_than_yaml() {
...
+    for file in "${OPT_SPEC%.spec}"{.spec,.yaml}; do
+        if [[ -f $file && $(file "$file") == *"CRLF line terminators"* ]]; then
+            crlf_used=1
+            break
+        fi
+    done
+
+    if [[ $crlf_used ]]; then
+        notice "Files with CRLF line endings found." \
+            "Consult the Sailfish SDK FAQ to learn why to avoid that and how."
```
vigejolla, line 820: Maybe provide link? And please remember to also add the information to the FAQ :)

vigejolla (approved):
LGTM

sdk-setup/src/mb2
martyone: I am afraid OMP might want to replace (rebrand) the link, so I rather used this free word reference.

vigejolla: They might want to do it anyway (i.e. change the text to say "Aurora SDK" instead of "Sailfish SDK"). Whether the text contains URL or not doesn't make much difference.

martyone: This is really different. While the name is a common value, so one system-wide mechanism is enough (e.g. /etc/os-release), there are many resources to link to, so something more would need to be done.

PR #205, opened 2020-10-20
[sdk-init] Install files with desired permissions. JB#50573

martyone:


sdk-setup/src/sdk-init
```diff
@@ -99,6 +99,30 @@ bad_usage()
 
 inside_build_engine() [[ -f /etc/mer-sdk-vbox ]]
 
+with_tmp_file()
+{
+    local file=$1 cmd=("${@:2}")
```
vigejolla, line 104: maybe $output_file instead of $file?

sdk-setup/src/sdk-init
```diff
@@ -99,6 +99,30 @@ bad_usage()
...
+            rm -f "$tmp_file"
+        fi
+    )
+    trap 'with_tmp_file_cleanup; trap - RETURN' RETURN
+    trap 'return 1' INT TERM HUP
+
+    tmp_file=$(mktemp "$file.XXX") || return
+
+    if "${cmd[@]}" <&3 >"$tmp_file"; then
+        cat <"$tmp_file" >"$file" || return
```
vigejolla, line 120: Maybe `cat "$tmp_file" > "$file" || return` instead?

sdk-setup/src/sdk-init
martyone: I kind of agree with the suggestion but OTOH this is a copy of a snippet that exists in multiple files already, so I would prefer keeping all occurrences in sync and only touch it if necessary :)

sdk-setup/src/sdk-init
martyone: ditto

vigejolla (approved):
LGTM, but I do wonder why `sed -i file` changes permissions while `cat  > file`doesn't.

sdk-setup/src/sdk-init
vigejolla: I didn't know that. I guess that makes sense.

martyone:
`sed -i` creates itself a temporary file with the same file mode as the original file has, then swaps those on success. But sed sees the original file through dynexecfs, so the file mode may not match what is on host. with_tmp_file does keeps the original file and only replaces its content.

PR #207, opened 2020-10-23
sdk-vm: Add simple stub to tell that the webapp is gone

Thaodan:


martyone (approved):

sdk-setup/systemd/sdk-webappstub.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Nothing to see here go away
+
+[Service]
+Type=simple
+ExecStart=/bin/sh -c "nc -l -p 8080 -N <<<$'HTTP/1.0 404 Not Found\n\nThe SDK WebApp is gone, use sfdk\n\n'"
```
vigejolla, line 6: As we are using bashism (<<<) here, I would be happier if we used /bin/bash instead of /bin/sh. Even though we now have /bin/sh pointing to /bin/bash, that's not something one should count on.

vigejolla: And please use "Control Centre" instead of SDK WebApp, as that's the term that was used in the SDK IDE.

vigejolla: And maybe "sfdk or Sailfish IDE"

vigejolla: It seems that at least firefox refuses to show the content unless the Content-length header is there as well

Thaodan:
Add the content-lenght header, reference to the control centre and the 
SailfishOS IDE.

sdk-setup/systemd/sdk-webappstub.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Nothing to see here go away
+
+[Service]
+Type=simple
+ExecStart=/bin/sh -c "while true ; do echo 'HTTP/1.0 404 Not Found\nContent-Length: 68\n\nThe Sailfish Control Centre is gone, use sfdk or the SailfishOS IDE.\n\n' | nc -l -p 8080;done"
```
martyone, line 6: Doesn't the echo require '-e'? ...hmmm "Sailfish SDK Control Centre" would be more correct.

PR #209, opened 2020-11-07
[mb2] Create .mb2 directory after all checks are done

Thaodan:
When running mb2 with an invalid option or in a directory without a
yaml/spec file mb2 still creates a .mb2 directory and litters the file
system.

### DIFF
--- a/sdk-setup/src/mb2
+++ b/sdk-setup/src/mb2
@@ -3098,8 +3098,6 @@ while [[ "$1" ]]; do
     esac
 done
 
-mkdir -p "$STATEDIR" || exit
-echo '*' > "$STATEDIR/.gitignore"
 
 if [[ -n "$OPT_SPEC" ]]; then
     try_to_make_spec "$OPT_SPEC"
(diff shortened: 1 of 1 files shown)


sdk-setup/src/mb2
```diff
@@ -3098,8 +3098,6 @@ while [[ "$1" ]]; do
     esac
 done
 
-mkdir -p "$STATEDIR" || exit
-echo '*' > "$STATEDIR/.gitignore"
```
vigejolla, line 3101: Extra empty line left here

vigejolla (changes requested):
There's no bug reference (JB#XXXXX) anywhere

sdk-setup/src/mb2
```diff
@@ -3184,6 +3182,9 @@ case "$1" in
         ;;
 esac
 
+mkdir -p "$STATEDIR" || exit
+echo '*' > "$STATEDIR/.gitignore"
+
```
martyone, line 3187: This is too late. At least the switch_to_snapshot call above needs $STATEDIR.

Thaodan:
Moved the check further up before `switch_to_snapshot` might be used.

martyone:
Pity the effect is really limited now :\ I think you could set up an exit trap and remove the .mb2 dir on exit if it contains nothing else than the pre-created .gitignore file.

Thaodan:
It is fine as the options are valid at this point.

martyone:
The .spec has not been processed yet at this point, so running mb2 with correct options but under a random directory with no .spec will leave .mb2 dir there.

Thaodan:
I moved the processing of the spec file a bit further just before the 
switching to the snapshot now.

martyone:
That's not possible either. `run_undeploy__process_args` and `run_check__process_args` may set `OPT_NEEDSPEC`, which would happen after spec processing (look for `${cmd}__process_args` invocation). I think the only workable option is to use the exit trap as suggested before.

Thaodan:
``` emacs-lisp

(defun test)
```

martyone:
Superseded by #282

PR #211, opened 2020-11-23
Use less to print long help.

jusa:
Initial motivation was to just change output from stderr to stdout so
that it is easier to grep the output but as other utils seem to utilize
less for long help change to that instead. This way it's still easy to
grep the help if needed but we use less by default.


martyone (approved):

martyone:
@jusa rebase please

jusa:
> @jusa rebase please

Done.

PR #215, opened 2020-12-09
Jb49109

martyone:


sdk-setup/src/mb2
```diff
@@ -1167,6 +1167,15 @@ get_device() {
     if [[ "$FOUND_DEVICE" == 1 ]]; then return 0; else return 1; fi
 }
 
+ensure_device_prepared() {
+    inside_build_engine || return 0
+
+    # Need to print reply in order to let it block
+    dbus-send --type=method_call --print-reply=literal --reply-timeout=30000 --system \
+        --dest="$SAILFISH_SDK_SFDK_DBUS_SERVICE" /device org.sailfishos.sfdk.Device.prepare \
+        >/dev/null || true
```
vigejolla, line 1176: Why are you using '|| true' here and '|| return 0' on line 1171? Isn't the purpose just to return successfully in both cases?

martyone: `|| return` is to return, `|| true` is to be explicit about the exit code being ignored intentionally and not by mistake.

vigejolla (approved):
LGTM

PR #216, opened 2020-12-16
[sdk-manage] Allow to remove targets with --force. JB#52253

vigejolla:
Removing targets regardless of snapshots or mode is needed for SDK
uninstallation.


sdk-setup/src/sdk-manage
```diff
@@ -1988,22 +1988,46 @@ reserve_target() {
 }
 
 remove_target() {
-    local target=$1
+    local force=
+    local target=
+    while [[ $# -gt 0 ]]; do
+        case $1 in
+            -f|--force)
+                force=true
```
martyone, line 1996: nitpick, `force=1` would be more in line with the existing code.

martyone (approved):
One nitpick worth to address.

PR #218, opened 2021-01-12
 [sdk-setup] Do not use predicable names for network interaces. Fixes JB#52706

martyone:
Predicable names are now used by default.


martyone:
I couldn't quickly figure out what triggered this behavior. This udev config was used already in the previous SDK release, I haven't found any config that could have been overriding that before.

vigejolla (approved):
The change itself looks good. I'm not so sure about the commit message. At least I understand the message so that after the commit, predictable names are used. Perhaps add one more sentence like "this commit restores the old behaviour" or something?

PR #222, opened 2021-02-10
Jb52875

martyone:


martyone:
Just the very last commit belongs to this PR.

vigejolla (approved):
LGTM. Even though I have to say that I spent way too much time trying to figure out what target_snapshot_needs_reset returns and when... But that's just mostly because bash is not a great programming language.

PR #223, opened 2021-02-11
Jb51836

martyone:


martyone:
Not properly tested yet but still ready for review.

sdk-setup/src/mb2
```diff
@@ -2326,58 +2326,57 @@ run_deploy() {
     [[ ! $all_opt || $method_opt != --rsync ]] || fatal "Cannot use '--all' with '--rsync'"
     [[ ${method_opt:-} ]] || fatal "Argument expected"
 
-    local retcode=0
+    # rsync method is special
+    if [[ $method_opt == --rsync ]]; then
+        ! behind_qtcreator || run_make_install || return
+        local name=$(get_spec_tag "%{name}")
+        rsync_as $device_username -av ${buildroot}/. $device_ip:/opt/sdk/$name
```
vigejolla, line 2341: Shouldn't we check return value and print "Deploy failed" here if needed?

sdk-setup/src/mb2
```diff
@@ -2326,58 +2326,57 @@ run_deploy() {
...
 
     case $method_opt in
         --sdk | --pkcon )
-            local method;
             if [[ $method_opt == --sdk ]]; then
-                method="sdk-deploy-rpm"
+                method=(sdk-deploy-rpm)
             else
-                method="pkcon --plain --noninteractive install-local"
+                method=(pkcon --plain --noninteractive install-local)
```
vigejolla, line 2367: Now this is just silly. Why don't you have separate cases for --sdk and --pkcon and get rid of the if?

sdk-setup/src/mb2
martyone: I was thinking if the message adds any value at all. In the original code it is only printed if it fails at the last step, while it may fail on various errors before that. Drop it?

But what is more important, `dry_run_opt` should be handled here :D

sdk-setup/src/mb2
martyone: Simply because I was just moving code around and facing other issues I didn't spot this one :)

sdk-setup/src/mb2
vigejolla: :-O Well I'm glad you reviewed your own code!

vigejolla (approved):
LGTM

sdk-setup/src/mb2
```diff
@@ -4022,10 +4022,31 @@ fix_version_tc() {
     (
         git() (
             echo "stub invoked: git" >&2
-            if [[ $1 == stash && $2 == create ]]; then
+
+            options=()
+            while [[ $# -gt 0 ]]; do
+                case $1 in
+                    -c)
```
vigejolla, line 4029: What's this? Where would the -c come from?

martyone: Check the commit referred from the commit message.

martyone: You will also find why I didn't hit this before :)

vigejolla: Right. I was clearly too tired on Friday as I didn't see it...

vigejolla (approved):
LGTM (again)

PR #225, opened 2021-02-16
Jb49209

martyone:


martyone:
Just the last two commits belong to this PR

jusa (approved):

sdk-setup/src/sdk-manage
```diff
@@ -1423,7 +1423,7 @@ synchronise_target() {
         maybe_times=
     fi
 
-    echo >&2 "Synchronising target to host"
+    echo >&2 "Synchronising target to host..."
```
vigejolla, line 1426: Have you considered adding --info=progress2 to rsync parameters?

martyone: That would be nice but unfortunately it does not print nicely with output not directed into terminal.

vigejolla (approved):
LGTM

PR #226, opened 2021-02-19
Jb42918

martyone:


vigejolla:
Do all of the commits belong to the PR? There are 20 of them...

vigejolla:
> Do all of the commits belong to the PR? There are 20 of them...

I guess I can at least ignore all sdk-host-shell related commits, seeing that it was dropped in the end

martyone:
Most of them were implemented before under other bugs that are now closed and the commits lived in a branch for long time, so I created just this single PR for all of them. Yes, sdk-host-shell is not needed nowadays, so you can ignore it. I think you can ignore the evolution of the code and simply review the final state.

sdk-setup/src/sdk-make-qmltypes
```diff
@@ -0,0 +1,1554 @@
...
+        return 1
+    fi
+
+    if ! behind_sfdk; then
+        fatal "This tool can be only used from session opened with sfdk."
+        return 1
+    fi
+
+    if [[ $OPT_BATCH ]]; then
+        if ! python3 -c 'import giturlparse' &>/dev/null; then
```
vigejolla, line 315: How big task would it be to package giturlparse, and make it a requirement?

sdk-setup/src/sdk-make-qmltypes
```diff
@@ -0,0 +1,1554 @@
...
+{
+    echo "Warning: $*" >&2
+}
+
+fatal()
+{
+    echo "Fatal: $*" >&2
+}
+
+configure()
```
vigejolla, line 302: I'm not sure this is the best name for the function. It doesn't actually configure anything, it just checks some prerequisites.

vigejolla:
> Most of them were implemented before under other bugs that are now closed and the commits lived in a branch for long time, so I created just this single PR for all of them. Yes, sdk-host-shell is not needed nowadays, so you can ignore it. I think you can ignore the evolution of the code and simply review the final state.

That's pretty much what I did

vigejolla (approved):
LGTM. Just some minor comments which you might consider. Packaging giturlrepo requires at least another task anyway.

sdk-setup/src/sdk-make-qmltypes
martyone: Small I believe. I just didn't pay much attention to this as it is only needed for the batch mode, which is not available (advertised) to regular users through sfdk.  I would package it later when utilizing the batch mode execution under sdk-test-suite.

sdk-setup/src/sdk-make-qmltypes
martyone: Checking for prerequisites is what configure scripts do first of all. Whether they can deal with particular prerequisite failing or not is just an additional level. Imagine it would install giturlparse instead of asking the user to do so and you wouldn't find it wrong.

martyone:
Rebased

sdk-setup/src/sdk-make-qmltypes
vigejolla: I didn't make the association with configure scripts. Sure, now that you put it like that, it makes sense.

sdk-setup/src/sdk-make-qmltypes
vigejolla: I'll add a task in bugzilla about that.

martyone: Thank you

martyone:
Rebased - polished commit/changelog messages.

PR #227, opened 2021-03-04
[mb2] Fix using sign packages without signing user. JB#50188

etaishev:


sdk-setup/src/mb2
```diff
@@ -1888,69 +1888,70 @@ sign_rpms_native() {
...
-    if [[ $OPT_PACKAGE_SIGN_USER ]]; then
-
-        if inside_build_engine; then
-            local shared_key=${OPT_SHARED_DIR}/gnupg/${OPT_PACKAGE_SIGN_USER}.key
-            if [[ ! -f $shared_key ]]; then
-                notice "Cannot sign packages: Internal error: No GPG key shared for the user '$OPT_PACKAGE_SIGN_USER'."
-                return 1
-            fi
+    if [[ ! $OPT_PACKAGE_SIGN_USER ]]; then
+        fatal "Cannot sign packages: No signing user specified."
```
martyone, line 1892: We used `notice` for all the others but now that I see this I think `fatal` would really be more accurate - if the user asks for signing, failure to do so should be fatal.

martyone, line 1892: It would be better to fail already at the time arguments are parsed, before the build starts. And not only in the case that signing user is not specified, but also in the other cases. Imagine the disappointment when build fails at the very end and it took 15 minutes to go there :)

So I would extract `sign_rpms_native__prepare` that would contain everything except the tmpdir creation and actual rpmsign call. There would be also `sign_rpms__prepare` which would do similar branching as the `sign_rpms` does now and this `sign_rpms__prepare` would be called from `run_build__process_args` and `run_package__process_args` if the `--sign` option is used.

sdk-setup/src/mb2
```diff
@@ -1856,104 +1856,107 @@ run_build_requires__process_args() {
     fi
 }
 
+sign_rpms__prepare() {
+    SIGN_SDKSIGN=/usr/libexec/sdk-setup/sdk-rpmsign
```
martyone, line 1860: This would be better set at file level here https://github.com/sailfishos/sdk-setup/blob/1.3.88/sdk-setup/src/mb2#L3032

sdk-setup/src/mb2
```diff
@@ -1856,104 +1856,107 @@ run_build_requires__process_args() {
...
-        $sdksign ${rpms}
+    if [[ -x $SIGN_SDKSIGN ]]; then
+        $SIGN_SDKSIGN ${rpms}
     else
         sign_rpms_native ${rpms}
     fi
 
     local rc=$?
-    [[ $rc -eq 0 ]] || { notice "Failed to sign packages"; return $rc; }
+    [[ $rc -eq 0 ]] || { fatal "Failed to sign packages"; }
```
martyone, line 1879: rc is unused now, let's check `$?` directly.

sdk-setup/src/mb2
```diff
@@ -1856,104 +1856,107 @@ run_build_requires__process_args() {
...
+    sign_rpms_native_cleanup() (
+        trap 'echo cleaning up...' INT TERM HUP
+        if [[ $tmpdir ]]; then
+            rm -rf "$tmpdir"
         fi
+    )
+    trap 'sign_rpms_native_cleanup; trap - RETURN' RETURN
+    trap 'return 1' INT TERM HUP
+
+    if [[ $SIGN_PASSPHRASE ]]; then
```
martyone, line 1951: I would feel a tiny bit more safer if the passphrase read from file was discarded immediately after it's use, so I would simply repeat this piece of code here instead of passing it by SIGN_PASSPHRASE:

```
    local passphrase=
    if [[ $OPT_PACKAGE_SIGN_PASSPHRASE_FILE ]]; then
            passphrase=$(<"$OPT_PACKAGE_SIGN_PASSPHRASE_FILE")
    else
            passphrase=$OPT_PACKAGE_SIGN_PASSPHRASE
    fi
```

sdk-setup/src/mb2
```diff
@@ -1856,104 +1856,107 @@ run_build_requires__process_args() {
...
-        if ! gpg2 --list-secret-keys "$OPT_PACKAGE_SIGN_USER" &>/dev/null; then
-            notice "Cannot sign packages: No GPG key found for the user '$OPT_PACKAGE_SIGN_USER'."
-            return 1
-        fi
+    if ! gpg2 --list-secret-keys "$OPT_PACKAGE_SIGN_USER" &>/dev/null; then
+        fatal "Cannot sign packages: No GPG key found for the user '$OPT_PACKAGE_SIGN_USER'."
+    fi
 
-        extra_args+=(--define "_gpg_name $OPT_PACKAGE_SIGN_USER")
+    SIGN_EXTRA_ARGS+=(--define "_gpg_name $OPT_PACKAGE_SIGN_USER")
```
martyone, line 1918: I would move this as a local variable to `sign_rpms_native` so that one can spot it all with one glance.

martyone (approved):

PR #230, opened 2021-03-19
[sdk-manage] Rebuild RPM db after target install. Fixes JB#53586

vigejolla:


martyone (commented):
I am a bit worried that we are hiding the issue. We could at least check first whether the db is broken (`rpm -qa` would fail?) and print a warning in that case, so that we have some reminder and if this is a temporary issue it might help us spot the root cause when it reappears.

vigejolla:
> I am a bit worried that we are hiding the issue. We could at least check first whether the db is broken (`rpm -qa` would fail?) and print a warning in that case, so that we have some reminder and if this is a temporary issue it might help us spot the root cause when it reappears.

rpm -qa does not fail. Any other ideas how to check?

martyone:
`rpm -e --test rpm`?

vigejolla:
> `rpm -e --test rpm`?

Nope, doesn't fail either. Well, it does fail, but it's because of dependencies. Tried with a package which actually could be removed (patterns-sailfish-device-configuration-Sailfish-SDK-Target) -> didn't fail. Dropping the "--test" makes it fail, but we really can't use that for the test because it would really remove the package if it happens to work :(

martyone:
yeah I forgot `--nodeps`, but if it didn't fail for the other package then I am getting out of quick ideas :\

martyone (approved):

PR #234, opened 2021-04-22
[sdk-init] Actually use -t parameter. Fixes JB#53853

vigejolla:


sdk-setup/src/sdk-init
```diff
@@ -125,7 +125,16 @@ with_tmp_file()
...
+    local wizard_file=
+    echo "Types:"
+    for template in "${DEF_TEMPLATES[@]}"; do
+        wizard_file="$OPT_SDK_ROOT/$DEF_TEMPLATES_PATH/sailfishos-$template/wizard.json"
+        template_displayname=$(grep -Po '"trDisplayName": "\K.*(?=")' $wizard_file|head -n 1)
+        echo "$template: $template_displayname"
+    done | column -t -s :
+
+    echo
     echo "    Builders: ${DEF_BUILDERS[*]} (Default: $DEF_BUILDER)"
```
martyone, line 138: Different templates may support different builders, so it should produce something like:

```
templateA ... Template A Description
    PropertyX: val1 val2
templateB ... Template B Description
    PropertyX: val2
```

Currently the QML-only template is qmake-only.

So we need to determine the available builders from the template, e.g. from:

```
"supportedProjectTypes": [ "CMakeProjectManager.CMakeProject", "Qt4ProjectManager.Qt4Project" ]
```

vigejolla: I didn't want to do a complete JSON parser :( I guess it's time to replace the grep with python...

martyone: A good portion of sdk-init has always been hacky, so I think simply grepping for "template.pro" and "CMakeLists.txt" in wizard.json won't hurt :)

BTW, having `jq` packaged would be handy :)

vigejolla: I already started with python, but with your permission I'll gladly throw it out and just do the grep you suggested :)

martyone (approved):
Some cosmetics issues worth to fix, otherwise LGTM.

PR #236, opened 2021-04-29
Jb54117

martyone:


sdk-setup/src/mb2
```diff
@@ -1032,8 +1032,13 @@ maybe_verify_target_dependencies() {
             filter_repo "$exclude_source_name" "$OPT_OUTPUTDIR" "$filtered_outputdir"
             local args=(--"$OPT_SEARCH_OUTPUTDIR_VERBOSITY" --non-interactive \
                 --plus-repo "$filtered_outputdir" --no-gpg-checks)
-            sb2 -t "$OPT_TARGET" -m sdk-install -R zypper "${args[@]}" dup --from '~plus-repo-1'
+            # Install first, then dup to ensure we really use packages from
+            # the output directory as the repos may contain newer packages and
+            # those would be pulled in preferably.  This happens not only when
+            # working with older sources but also when a package hardcodes
+            # version in its spec and the --fix-version option is not active.
```
vigejolla, line 1039: Now this seems so logical that I'm starting to wonder if there was some reason (that I'm not aware of and you have already forgotten) why dup was originally performed before install?

martyone: Yeah, seems logical but it feels more natural/safe to dup first and install new packages later - this is what you would normally do to your system, right? In theory, ordered install-first it could fail temporarily and there is the possibility to get a package pulled from a regular repo first and downgraded to the local built version in a second step - harmless but slightly annoying.

vigejolla (approved):
I now understand (at least I think so ;) both why we had the order the other way around originally, and why this now fixes the issue. I still think having the possibility to skip the dup entirely wouldn't hurt though...

martyone:
There is a demand for another feature we discovered yesterday: optionally preinstall additional (not build-required) packages into a *temporary* snapshot. Optionally == do not fail when they do not exist as they may be provided locally later. Instead of making this a built-in feature of mb2, I would like to add a generic mechanism that would allow freely tweaking the temporary snapshots. This would also allow completely overriding the way build requires all pulled in.

PR #237, opened 2021-05-05
Jb54141

martyone:


sdk-setup/src/mb2
```diff
@@ -1029,27 +1182,43 @@ maybe_verify_target_dependencies() {
...
     fi
 
-    if [[ $OPT_SEARCH_OUTPUTDIR ]]; then
-        # Install first, then dup to ensure we really use packages from
-        # the output directory as the repos may contain newer packages and
-        # those would be pulled in preferably.  This happens not only when
-        # working with older sources but also when a package hardcodes
-        # version in its spec and the --fix-version option is not active.
-        sb2 -t "$OPT_TARGET" -m sdk-install -R zypper "${args[@]}" dup --from '~plus-repo-1'
+    run_hook pre-pull-build-requires "${hook_exports[*]}" "${deps[@]}" \
```
vigejolla, line 1198: Shouldn't we check with has_hook pre-pull-build-requires first?

vigejolla: right, it's checked inside run_hook, nevermind

vigejolla (approved):
LGTM. Interesting approach, which provides lot's of opportunities for custom tweaking. I like it :)

PR #240, opened 2021-05-19
Jb54305

martyone:


vigejolla (approved):
LGTM. But it leaves me wondering why did the old behaviour result in the error?

martyone:
With the original approach the host-side code could temporarily see either empty or incomplete content.

vigejolla:
> With the original approach the host-side code could temporarily see either empty or incomplete content.

Yes, I understand that, but it just seem unlikely that the host-side code would want to read the content at that precise moment. Is it watching for changes in the file?

martyone:
Yup, it watches it.

PR #241, opened 2021-05-28
Jb54236

martyone:


vigejolla (approved):
LGTM

sdk-setup/src/sdk-manage
```diff
@@ -2046,6 +2046,9 @@ reserve_target() {
 
         pooled_targets+=("$target")
 
+        # Target currently in use
+        [[ $other_lock_file == $lock_file ]] || ! is_locked "$other_lock_file" || continue
```
vigejolla, line 2050: Shouldn't we instead print some kind of error message and exit?

martyone: Nope, we are looking for a suitable target from a pool. It's expected that some are not suitable, e.g., because they are in use by other process.

vigejolla: Yes, of course. For some reason I was confused and thought we wanted to use a specific target.

PR #246, opened 2021-06-17
Jb54046

martyone:


martyone:
Just the last commit belongs to this PR

sdk-setup/src/sdk-manage
```diff
@@ -1469,6 +1469,7 @@ synchronise_target() {
 # Qt5 qml imports are here
 + $libdir/qt5/qml**
 + /usr/include**
++ $libdir/*/include**
```
vigejolla, line 1472: What about the /usr/share dirs which were mentioned in the bug?

martyone: Those from cmake are already covered, those from libtool are not relevant IIUC.

vigejolla (approved):
LGTM

sdk-setup/src/sdk-manage
vigejolla: Actually, cmake is *not* covered as far as I can tell: /usr/lib/cmake is synced, but /usr/share/cmake/Modules and /usr/share/cmake/Modules and /usr/share/cmake/include, which were listed in the bug, are not.

martyone: Indeed :( and it's the commit on the very bottom... aargh

martyone: I will push another commit directly to master after merging all the PRs.

martyone: https://github.com/sailfishos/sdk-setup/commit/57ec7b54c62522b42b2257be16ad6f45d2f494ae

PR #248, opened 2021-06-18
Jb54454

martyone:


martyone:
Just the last three commits belong to this PR.

sdk-setup/src/mb2
```diff
@@ -415,29 +416,31 @@ COMMANDS
...
 
         --rsync
-            This method is special.  Uses 'rsync' to copy the installed files
-            under '/opt/sdk/<name>' without building/installing RPM package(s).
-            Does not require root access to the device.  Incompatible with
-            '--all'.
+            This method is special.  It can be used to transfer the RPM
+            "buildroot" to the device as '/opt/sdk/<name>' without building
+            and installing RPM packages (see also 'make-install').
+            Incompatible with '--all'.
```
vigejolla, line 426: I think the wording could still be improved. This can be interpreted as meaning "can be used without the need to build and install" while quite the opposite is true. It's not a big deal, as the command will now print a warning if buildroot is empty, so the user should realize his mistake at that point anyway.

vigejolla (approved):
LGTM. Wrote one comment regarding the help text, but as I don't really have any better suggestion for the text, feel free to ignore it unless you come up with something.

sdk-setup/src/mb2
martyone: yeah that's because there are at least three context for "build and install" involved. I will think about improving it the next round - no idea at hand.

martyone: handled in PR #500

PR #253, opened 2021-07-07
[sdk-setup] Remove jolla-ca zypper requires. JB#54898

etaishev:


martyone:
Simply dropping it doesn't look right to me. Shouldn't we replace it with some more generic capability?

etaishev:
I can't see all Jolla code, but as i understand from PRs in jb49356 - jolla-ca will be in sdk any way

vigejolla:
> I can't see all Jolla code, but as i understand from PRs in jb49356 - jolla-ca will be in sdk any way

The point of the the check files is that a warning will be displayed if the user tries to remove the package. So instead of just removing jolla-ca, we should have jolla-ca provide some more generic capability, let's say 'vendor-ca'. And then have the check require that instead.

etaishev:
Update PR. Add check for vendor-ca. 
Who can update 'jolla-ca' for `Provides: vendor-ca`?

okodron (approved):

etaishev:
Update on current master

vigejolla (approved):
typo in the commit message (virual -> virtual). Otherwise lgtm.

etaishev:
@martyone Can you please see

stanislavmaksimov:
@martyone, @vigejolla, could you please have a look

PR #255, opened 2021-07-16
[sdk-vm] Add sdk-freespace service. JB#54555

vigejolla:


sdk-setup/src/sdk-freespace
```diff
@@ -0,0 +1,28 @@
...
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE 
+# DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY 
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES 
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; 
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND 
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT 
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS 
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+df --output=avail / | tail -n 1 > /host_targets/df.cache
```
martyone, line 28: I would prefer `/etc/mersdk/share/df.cache`.

rpm/sdk-setup.spec
```diff
@@ -302,6 +305,7 @@ fi
 %{_bindir}/sdk-setup-enginelan
 %{_bindir}/sdk-shutdown
 %{_bindir}/dynexecfs
+%{_bindir}/sdk-freespace
```
martyone, line 308: I would prefer hiding it from PATH, so using `%{_libexecdir}/%{name}` as a prefix. Maybe we could even omit the executable and specify the oneliner literally in the .spec file as in `ExecStop=/bin/bash -c 'df ...'`?

sdk-setup/src/sdk-freespace
vigejolla: That was my first thought as well, but as /etc/mersdk/share is not writable, I thought host_targets would do. Of we can change /etc/mersdk/share to be writable, if that's what you really want?

rpm/sdk-setup.spec
vigejolla: Right, specifying it directly on the service file makes sense as long as it stays oneliner.

sdk-setup/src/sdk-freespace
martyone: hm, I wasn't sure about it being RW (and I couldn't check from inside the build engine at that moment), so I checked just the sources https://github.com/sailfishos/sdk-setup/blob/master/sdk-setup/systemd/etc-mersdk-share.service#L22 before I suggested that. Now I see the installer configures the VBox share as RO, so that explains it. Yup, I think it make sense to let it be RW.

martyone (approved):

sdk-setup/systemd/sdk-freespace.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Update df.cache at shutdown
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStop=/bin/bash -c 'df --output=avail -B1048576 / | tail -n 1 | sed "s/^ *//" > /etc/mersdk/share/df.cache'
```
martyone, line 7:
```suggestion
ExecStop=/bin/bash -c 'df --output=avail -B1048576 / |sed -n "$s/^ *//p" > /etc/mersdk/share/df.cache'
```

PR #265, opened 2021-10-08
Jb55389

martyone:


sdk-setup/src/mb2
```diff
@@ -2739,11 +2776,58 @@ run_package() {
...
+}
+
+sb2_maintain() {
+    local sb2_args=(-m sdk-install -R)
+    if [[ $OPT_BUILD_SHELL_MAINTENANCE_ARGS ]]; then
+        read -a sb2_args <<<$OPT_BUILD_SHELL_MAINTENANCE_ARGS || return
+    fi
+
+    if [[ $# -gt 0 ]]; then
+        sb2 -t $OPT_TARGET "${sb2_args[@]}" -- "$@"
```
vigejolla, line 2827: Why is "-t $OPT_TARGET" not part of sb2_args? It seems a bit random to handle it separately.

vigejolla: For some reason I somehow thought `read -a sb2_args <<<$OPT_BUILD_SHELL_MAINTENANCE_ARGS` would append to sb2_args, instead of replacing existing values. I blame lack of coffee.

PR #266, opened 2021-10-13
[packaging] BuildRequire systemd via pkgconfig. JB#55010

Thaodan:
This allows to pick systemd-mini inside the builder instead of full
systemd.

Signed-off-by: Björn Bidar <bjorn.bidar@jolla.com>

### DIFF
--- a/rpm/sdk-setup.spec
+++ b/rpm/sdk-setup.spec
@@ -1,14 +1,14 @@
 Name:       sdk-setup
 
 Summary:    SDK setup packages for Mer SDK
-Version:    0.63
+Version:    1.4.14
 Release:    1
 Group:      System/Base
 License:    GPL
 BuildArch:  noarch
 URL:        https://github.com/sailfishos/sdk-setup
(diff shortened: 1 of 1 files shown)


rpm/sdk-setup.spec
```diff
@@ -8,7 +8,7 @@ License:    GPL
 BuildArch:  noarch
 URL:        https://github.com/sailfishos/sdk-setup
 Source0:    %{name}-%{version}.tar.bz2
-BuildRequires:  systemd
+BuildRequires:  pkgconfig(systemd)
```
martyone, line 11: IIUC systemd is build required here just for its RPM macros like `%systemd_requires`, `%systemd_post` etc. It makes no use of pkgconfig.

So this makes me think:

1. Are those macros possibly packaged separately nowadays? If not, then
2. Is this change really required for all packages, including sdk-setup, or just those that may come into way at bootstrap time? If yes, then
3. Isn't there an alternative way to let it choose between systemd and systemd-mini, more suitable for cases like this?

PR #272, opened 2021-10-29
[sdk-manage] Use rpmdb directly. Fixes JB#56088

vigejolla:


martyone (approved):
One suggestion, otherwise LGTM

sdk-setup/src/sdk-manage
```diff
@@ -692,15 +692,15 @@ rebuild_rpmdb() {
 
     # rpm can't overwrite directory on a different overlay layer
     if [[ $(findmnt --noheadings --output FSTYPE / ) == overlay ]]; then
-        enter "$object" rpm --rebuilddb 2>&1 | sed -e '/failed to replace old database/,+1d'
+        enter "$object" rpmdb --rebuilddb 2>&1 | sed -e '/failed to replace old database/,+1d'
```
martyone, line 695: I would add a comment here e.g. "Use rpmdb directly to work around an sb2 bug"

PR #273, opened 2021-11-08
[mb2] Exit if applying patch fails. JB#56248

vigejolla:


sdk-setup/src/mb2
```diff
@@ -2315,7 +2315,8 @@ run_apply() {
             fi
             patch_file="$(sed -n "s/^Patch$p: \+\(.*\)/\1/p" "$TMP_SPEC")"
             echo "$message $patch_file" >&2
-            patch "${common_op[@]}" $op < "$(dirname "$OPT_SPEC")/$patch_file" >&2
+            patch "${common_op[@]}" $op < "$(dirname "$OPT_SPEC")/$patch_file" >&2 \
+                || fatal "Applying $patch_file failed, exiting"
```
martyone, line 2319: It may be applying but it may be reversing as well (see `$message`). I'd consider passing the patch file through the `--input` switch instead of stdin (in the hope that `patch` would mention the file name in its error output) and make the `fatal` message more generic, like "patch failed".

martyone: BTW, is the `>&2` redirection really needed?

vigejolla: Changed the message to use $message- Using --input made no difference in patch output.

vigejolla: > BTW, is the `>&2` redirection really needed?

You should know, you added it with commit message "mb2: Do not report progress on stdout"

martyone: Yeah `patch` really reports progress on stdout

martyone (approved):

PR #275, opened 2021-11-12
Jb49237

martyone:


sdk-setup/src/sdk-make-qmltypes
```diff
@@ -865,6 +873,33 @@ revert_removal_of_manual_additions()
...
 }
 
+find_installed_revision()
+{
+    [[ $1 = self ]] || local -n self=$1
+    local repo=$2
+    local prefix=${3:-v}
+
+    local main_binary=
+    main_binary=$(job_expand_binaries self |awk '{print $1}') || return
```
vigejolla, line 883: Isn't this actually the package name? The word "binary" is usually used for the executables, so this is a bit confusing.

vigejolla (approved):
LGTM; only one minor nitpick about a variable name

sdk-setup/src/sdk-make-qmltypes
martyone: Yeah it's confusing when you miss the context. In this source the word "binary" always means a "binary (RPM) package". Compare with e.g. man rpmbuild.

martyone: maybe you missed "main_**binary**=$(job_expand_**binaries** ...)"

martyone: but that would not likely help you anyway as `job_expand_binaries` may mean many things as well :D

vigejolla: > Yeah it's confusing when you miss the context. In this source the word "binary" always means a "binary (RPM) package". Compare with e.g. man rpmbuild.

Perhaps main_binary_package would then be better name? Or perhaps just call it "b" and use that for all variables which deal with binary packages? Then you can also say that the reader should be able to understand that in this context "b" always means "binary package".

> maybe you missed "main_**binary**=$(job_expand_**binaries** ...)"

No, I did not. I would have complained about job_expand_binaries as well if it had been part of this PR.

martyone: Well, I feel it quite balanced WRT to 100 chars wrap margin. One can hardly do anything in the real world not implying some scope. Consistency is more important.

PR #278, opened 2021-12-03
Jb48346

martyone:


sdk-setup/src/mb2
```diff
@@ -964,6 +964,61 @@ make_gvariant_array() {
...
+%{nil}'
+
+    # Let rpmbuild acquire the lock at the beginning of the %install section
+    # (before the shared build root or a shared output directory is accessed)
+    EXTENDED_PRE_INSTALL_LOCK_ARGS=(--define "__spec_install_pre \
+$def_spec_install_pre_head\\
+flock ${GLOBAL_LOCK_INTERCHANGE_LOCK@Q} flock ${GLOBAL_LOCK@Q} \\\\\\
+    ln -f ${GLOBAL_LOCK_EXTENDED_LOCK@Q} ${GLOBAL_LOCK@Q}\\
+$def_spec_install_pre_tail")
+}
```
vigejolla, line 1004: This is a bit difficult to follow, but I guess that can't be really helped; A diagram would help a lot, but of course that's a bit difficult to include in a comment. But please verify that I understood this correctly: The aim is that that once the build process reaches the install step, the already locked extended lock will replace the global lock, and it will only be freed after mb2 exits. Meaning that if we have multiple build processes running concurrently, they will happily continue, possibly locking and freeing the global lock as they go, but once one of the processes reaches the install step, the other processes will wait at the next global_lock_acquire() or beginning of install step until the first process has finished completely. I.e. After the lock has been interchanged, global_lock will remain locked until the mb2 process which locked it has been finished.

vigejolla (approved):
LGTM, assuming I understood the principle correctly :)

sdk-setup/src/mb2
martyone: Yup, this ensures that no other build will mess up the RPM build root between %install and the actual package creation, which is done by various processes spawned by rpmbuild and by rpmbuild itself. So the lock is owned by sfdk but only comes into the play with the %install section.

martyone:
Small update on error handling ^

martyone:
Fixed self-test ^

vigejolla:
still LGTM

PR #283, opened 2021-12-08
Jb56649

martyone:


martyone:
Just the last three commits belong to this PR

sdk-setup/src/mb2
```diff
@@ -731,6 +731,9 @@ GLOBAL OPTIONS
         packages, read the passphrase from the <file>. Only the first line will
         be read.
 
+    --package-timeline
```
vigejolla, line 732: What's the story behind the name? Why not '--preserve-old-packages' ?

vigejolla (approved):
I'd like to get an answer to the one question, otherwise LGTM

sdk-setup/src/mb2
martyone: In order to group well, the name should begin with "package" (`--foo-bar` is often represented as `foo.bar` on sfdk side), but `--package-preserve-old` would not read well. Hence the option to get a timeline of your packaged work, which fits well between the other `package.*` options.

PR #289, opened 2021-12-15
Jb53186

martyone:


vigejolla:
LGTM so far

sdk-setup/src/proxymanager
```diff
@@ -0,0 +1,50 @@
...
+DBusGMainLoop(set_as_default=True)
+
+bus = dbus.SystemBus()
+manager_object = bus.get_object('org.pacrunner', '/org/pacrunner/manager')
+manager = dbus.Interface(manager_object, dbus_interface='org.pacrunner.Manager')
+
+with open(CONFIG_FILE, 'r') as proxy_json:
+    config = json.load(proxy_json)
+    manager.CreateProxyConfiguration(config)
+
```
vigejolla, line 47: If I understood this correctly, we only read the config file on startup? Maybe we could also watch it for changes?

martyone: With the original configuration we did it on startup and then on every connect. But I think it's better to avoid changes while the engine is (potentially) in use.

vigejolla (approved):
LGTM

PR #290, opened 2021-12-15
[sdk-chroot] Rename mer-sdk-chroot to sdk-chroot. Fixes JB#56734

vigejolla:
The script still has several references to mer. Those will be handled in separate tasks.

### DIFF
--- a/sdk-setup/src/sdk-chroot
+++ b/sdk-setup/src/sdk-chroot
@@ -1,10 +1,5 @@
 #!/bin/bash
-# mer-sdk-chroot
-
-# TODO
-#
-# Support a Mer clean setup (ie no .oscrc)
-# Support multiple shells (ie split setup/entry)
+# sdk-chroot
 
 usage()
 {
@@ -12,8 +7,8 @@ usage()
     usage: $0 [-u <user>] [-m <all|none|root|home>] [-r <SDK root path>] [<command> <args> ..]
            $0 -h
 
-       This is the Mer chroot SDK.
-       For information see https://sailfishos.org/wiki/Platform_SDK
+       This is the minimal, chroot-based Sailfish SDK.
+       For information see https://docs.sailfishos.org/Tools/Platform_SDK/
 
       If command is not present,
          used to enter the SDK and begin working. The SDK bash shell is a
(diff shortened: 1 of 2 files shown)


martyone:
I would prefer just "sdk-chroot".

vigejolla:
> I would prefer just "sdk-chroot".

Fine by me.

sdk-setup/src/sdk-chroot
```diff
@@ -1,19 +1,14 @@
...
 
 usage()
 {
     cat <<EOF
     usage: $0 [-u <user>] [-m <all|none|root|home>] [-r <SDK root path>] [<command> <args> ..]
            $0 -h
 
-       This is the Mer chroot SDK.
-       For information see https://sailfishos.org/wiki/Platform_SDK
+       This is the Sailfish OS chroot SDK.
```
martyone, line 10: maybe

```suggestion
       This is the minimal, chroot-based Sailfish SDK.
```

martyone (approved):
One suggestion

sdk-setup/src/sdk-chroot
Thaodan: Also don't remove the link to the wiki.

martyone:
Provide an alias with deprecation warning for convenience?

martyone:
And PR title needs update.

Thaodan:
> I would prefer just "sdk-chroot".

There's more than one sdk so sdk ambiguous imho especially if you put the script into path.

sdk-setup/src/sdk-chroot
vigejolla: > Also don't remove the link to the wiki.

I didn't remove the link. I replaced it with link to the new documentation site.

vigejolla:
> There's more than one sdk so sdk ambiguous imho especially if you put the script into path.

We don't recommend putting the script to path, we recommend using an alias.

vigejolla:
> Provide an alias with deprecation warning for convenience?

Sure, why not

Thaodan:
> @vigejolla commented on this pull request.
> 
> 
> 
> 
>  usage()
>  {
>      cat <<EOF
>      usage: $0 [-u <user>] [-m <all|none|root|home>] [-r <SDK root path>]
> [<command> <args> ..] $0 -h
> 
> -       This is the Mer chroot SDK.
> -       For information see https://sailfishos.org/wiki/Platform_SDK
> +       This is the Sailfish OS chroot SDK.
> 
> > Also don't remove the link to the wiki.
> 
> I didn't remove the link. I replaced it with link to the new documentation
> site.
> 
Ah ok the diff was shown wrong then.

vigejolla:
> Ah ok the diff was shown wrong then.

You didn't look at the whole diff, you looked only at the small section which was shown attached to martyone's comment. As martyone's comment was on line just above the new link, the new link was not show.

martyone:
I was rather thinking of `ln -s sdk-chroot mer-sdk-chroot` and then `[[ $0 = *mer-sdk-chroot ]] && echo ...` inside sdk-chroot. ...so that possible scripts do not break for users immediately.

vigejolla:
> I was rather thinking of `ln -s sdk-chroot mer-sdk-chroot` and then `[[ $0 = *mer-sdk-chroot ]] && echo ...` inside sdk-chroot. ...so that possible scripts do not break for users immediately.

I thought about that, but then I realized the users would quite likely just miss the warning and never update their scripts. It's better to fail fast with an error message.

martyone (approved):
LGTM, quoting the message would not hurt ;)

PR #294, opened 2021-12-23
[mb2] Fix buildroot timestamp check under VBox. Fixes JB#56804

vigejolla:
For some reason comparing timestamps between vboxsf and other
filesystems does not work properly.

### DIFF
--- a/sdk-setup/src/mb2
+++ b/sdk-setup/src/mb2
@@ -2892,7 +2892,7 @@ run_make_install() {
     ) || return
 
     mkdir -p "$buildroot" # hack for self-test to pass with stubbed sb2
-    touch --reference="$buildroot" "$INSTALL_STAMP"
+    stat -c %Y "$buildroot" > "$INSTALL_STAMP"
 }
 
 run_package__process_args() {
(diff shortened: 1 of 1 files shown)


sdk-setup/src/mb2
```diff
@@ -3071,7 +3072,10 @@ run_deploy() {
             fatal "Nothing to deploy. Maybe you forgot to use the \"make-install\" command."
         fi
 
-        if [[ -e $INSTALL_STAMP && $buildroot -nt $INSTALL_STAMP ]]; then
+        if [[ -e $TMP_INSTALL_STAMP ]]; then
+            [[ $buildroot -nt $TMP_INSTALL_STAMP ]] && \
+                fatal "The content of the RPM \"%buildroot\" directory does not match this build."
+        elif [[ -e $INSTALL_STAMP && $buildroot -nt $INSTALL_STAMP ]]; then
```
martyone, line 3078: We could better store the timestamp as file data in run_make_install

```
stat -c %Y "$buildroot" > "$INSTALL_STAMP"
```

and then

```suggestion
        if [[ -e $INSTALL_STAMP && $(stat -c %Y "$buildroot") -ne $(<"$INSTALL_STAMP") ]]; then
```

vigejolla: Right. I'll test with that.

vigejolla: Seems to work

martyone (approved):
PR summary needs update. Also, maybe say "Fix buildroot timestamp check under VBox." instead?

PR #295, opened 2022-01-11
[mb2] Disable STATEDIR check for build-shell. Fixes JB#56921

vigejolla:


sdk-setup/src/mb2
```diff
@@ -3935,7 +3935,7 @@ case $1 in
         mkdir -p "$STATEDIR" || exit
         echo '*' > "$STATEDIR/.gitignore"
         ;;
-    check | run | ssh )
+    check | run | ssh | build-shell )
```
martyone, line 3938: Even build-shell needs STATEDIR. It stores at least build target snapshot lock there. So it is inevitable to init build directory prior to using this command.

vigejolla: Does it? Where is that build target snapshot locking implemented, I don't see it. Also, what does it protect from?

martyone: With snapshot pool used, locking prevents other builds from hijacking a snapshot while it is in use. https://github.com/sailfishos/sdk-setup/blob/a2b2511ce6a00befc469ab2d217561625e59079b/sdk-setup/src/mb2#L2265.

vigejolla: But that is only used with pooled snapshots - it's not used with the default snapshot. Should also the default snapshot be locked when used?

martyone: This locking is only dealing with protection of pooled snapshots from being reset/removed while used. Dealing with simultaneous use of a build target is handled at more fine grained level independently.

vigejolla: Still, to me it looks like just executing "sfdk build-shell" does not lock anything. I can run multiple sfdk build-shells for the same target simultaneously. So your statement that the STATEDIR is needed for locking seems false.

martyone: Try with `sfdk config snapshot=%pool,2`, then use `sfdk build-shell` in three separate build dirs simultaneously. You should hit https://github.com/sailfishos/sdk-setup/blob/1ad8f4cd6a6989a89f0b327e82092ad8d422309a/sdk-setup/src/sdk-manage#L2138.

vigejolla: I'm talking about situation when using the default snapshot. Which is what most users probably use, as it is the default.

martyone: I am not sure what you mean? :) Breaking non-default use cases is OK?

Note that I do not find it wrong that the command needs to be used from (the top of) an initialized build directory. It is consistent with other build commands, it allows us to expand the state information we work with in future without further changes, and it also limits the space for issues - mb2 was always meant to be used from the top of a build dir, it was just not explicit about and people had to learn the hard way. It's great that it is clear and loud about it now.

vigejolla: > Breaking non-default use cases is OK?

No, of course not.

> Note that I do not find it wrong that the command needs to be used from (the top of) an initialized build directory.

Yes, I can see that. I'm just trying to find out why that is. So the reason is that there are use cases where things get broken - it just so happens that the cases which the automated tests test are not among those. Because of that it feels a bit weird to me to arbitrarily require them to be run from initialized build directory, just because some other use cases could be broken.

Please note that I'm not trying to argue about anything here, I'm just trying to understand your reasoning. I'm fine with changing the tests so that they run build-init first, I just want to understand why I'm doing that instead of changing the implementation so that it's not required.

martyone: > just because some other use cases could be broken.

The less exceptions the better. Both for easier understanding to the current behavior and for expanding the behavior in future.

One example for future extensions: we still miss the check to prevent accidental use of different build target on existing build directory when you come back later. Annoying when you break build that takes 5 minutes to redo. Remember people fighting strange build errors, finding out that they didn't fully clean up prior to switching to other build target. We should always record the target name under the STATEDIR during initialization and refuse to use it with other targets.

vigejolla: > One example for future extensions: we still miss the check to prevent accidental use of different build target on existing build directory when you come back later. Annoying when you break build that takes 5 minutes to redo. Remember people fighting strange build errors, finding out that they didn't fully clean up prior to switching to other build target. We should always record the target name under the STATEDIR during initialization and refuse to use it with other targets.

Yeah, we should do this. I have done that a few times myself.

vigejolla:
Superseded by https://github.com/sailfishos/sdk-test-suite/pull/16

PR #300, opened 2022-01-18
Jb57092

martyone:


sdk-setup/src/sdk-manage
```diff
@@ -545,6 +545,13 @@ object_config_get()
     echo "${value:-$default}"
 }
 
+object_config_get_boolean()
+{
+    local value=
+    value=$(object_config_get "$@") || return
+    tr -d '0' <<<$value
+}
```
vigejolla, line 560: This feels a bit like a hack. But it does work, and it is convenient. And with bash everything feels like a hack. So I guess it's good :)

martyone: Yeah after all I did it a more traditional way :)

martyone: LOL, I was thinking too fast. Reverted to the hacky way as that allows to separate error handling :)

vigejolla (approved):
LGTM

PR #302, opened 2022-01-20
[sdk-chroot] Add support for usernames with a dot.

georgyVasilev:


sdk-setup/src/sdk-chroot
```diff
@@ -294,6 +294,10 @@ prepare_user() {
 
     echo "$user ALL=NOPASSWD: ALL" > ${sdkroot}/etc/sudoers.d/$user
     chmod 0440 ${sdkroot}/etc/sudoers.d/$user
+    
+    if [[ $user == *['!'.]* ]] ; then
```
vigejolla, line 298: What's the '!' there for? Wouldn't just `*[.]*` suffice?

sdk-setup/src/sdk-chroot
```diff
@@ -294,6 +294,10 @@ prepare_user() {
...
+        mv ${sdkroot}/etc/sudoers.d/$user ${sdkroot}/etc/sudoers.d/${user//./}
```
martyone, line 299: How about to simply not use `$user` here, i.e., use a static name for the file, like "sdk"?

vigejolla: That's like... too simple :) It would also work with a user named '01_keep_zypp_logfile' :)

sdk-setup/src/sdk-chroot
georgyVasilev: Yes, it seems so

vigejolla (approved):
LGTM

PR #304, opened 2022-01-26
Jb57205

martyone:


martyone:
Unfortunately this approach does not work as the core.filemode option is often (always?) set at local scope during repository initialization.

PR #306, opened 2022-01-27
Jb57205

martyone:


vigejolla (approved):
Perhaps the comment should mention that the first "stash create" touches the index and that's why it works on the second run. I'm not sure if I can say LGTM, but approving anyway.

sdk-setup/src/mb2
```diff
@@ -1526,8 +1526,9 @@ maybe_fix_package_version() {
     version=${version#*/}   # allow tags to have a prefix to allow vendor marking
     version=${version#v}    # some people like to prefix versions with a v
 
+    # Git stash is known to fail on git index which was updated on host previously
     local stash=
-    if ! stash=$(git_ stash create); then
+    if ! stash=$(git_ update-index --refresh >/dev/null && git_ stash create); then
```
vigejolla, line 1531: Doesn't this fail when there are modifications?

martyone: How obvious :'-( ...fixed.

vigejolla (approved):
LGTM

PR #310, opened 2022-01-28
Jb57230

martyone:


martyone:
Just the very last commit belongs to this PR

sdk-setup/src/sdk-manage
```diff
@@ -489,12 +489,13 @@ get_all_or_selected_of_type() {
...
     fi
 
-    # Exclude snapshots
+    # Exclude snapshots unless required explicitly
     if [[ $type == target ]]; then
         local objects_=("${objects[@]}")
         objects=()
         for object in "${objects_[@]}"; do
-            target_is_snapshot "$object" || objects+=("$object")
+            ! target_is_snapshot "$object" || ! [[ $get_all ]] || continue
```
vigejolla, line 497: This makes my brain hurt. But I *think* the logic is correct.

vigejolla (approved):
LGTM

sdk-setup/src/sdk-manage
martyone: Whenever you see `!A || B`, read it as `A -> B` :)

vigejolla: Yes, I know. Still having two of those in a row is too much for me, at least on a friday afternoon :)

PR #313, opened 2022-02-10
[mb2] Run rpmlint by default after build. JB#57397

vigejolla:


vigejolla:
At this time this is meant only for platform developers. I'm thinking of modifying sfdk so that it will pass --no-rpmlint so sfdk users won't be affected.

jusa (approved):

rainemak:
Please rebase on top of master before merging.

rainemak (approved):

sdk-setup/src/mb2
```diff
@@ -3515,7 +3527,7 @@ run_check() {
     trap 'run_check_cleanup; trap - RETURN' RETURN
     trap 'return 1' INT TERM HUP
 
-    global_lock_acquire
+    [[ $no_lock ]] || global_lock_acquire
```
martyone, line 3530: There is an unpaired `global_lock_release` in `run_check_cleanup` now.

sdk-setup/src/mb2
```diff
@@ -3482,6 +3489,11 @@ run_check__process_args_helper() {
 }
 
 run_check() {
+    local no_lock=
+    if [[ $1 == --no-lock ]]; then
```
martyone, line 3493: I would prefer naming it `--already-locked` - it took me a while to get it.

sdk-setup/src/mb2
```diff
@@ -3587,6 +3599,14 @@ run_check() {
     return $rc
 }
 
+maybe_run_check() {
+    if [[ ! $BUILD_NORPMLINT ]]; then
```
martyone, line 3603: The use of `BUILD_*` variable to control the behavior implies that the function is only used from `run_build`, so given that it is just few lines long, I would suggest to implement it directly under `run_build` instead.

sdk-setup/src/mb2
```diff
@@ -2650,6 +2655,8 @@ run_build() {
     local hook_exports=()
     local rpms=$(find_rpms "$OPT_OUTPUTDIR" "$OPT_PKGDIR" "$STATEDIR/spec")
     run_hook post-package "${hook_exports[*]}" ${rpms} || fatal "The 'post-package' hook failed"
+
+    maybe_run_check
```
martyone, line 2659: Now rpmlint failure results to non-zero overall exit code. Shouldn't we relax it?

sdk-setup/src/mb2
```diff
@@ -2573,6 +2575,9 @@ run_build__process_args() {
             --no-check) shift
                 BUILD_NOCHECK_ARGS=(--nocheck)
                 ;;
+            --no-rpmlint) shift
+                BUILD_NORPMLINT=1
```
martyone, line 2579: Can we go without the option and simply decide with `BUILD_NORPMLINT=${SAILFISH_SDK_FRONTENT:+1}`? There should be little need to use `--no-rpmlint` on user side and it would also remove the need to adapt on sfdk side.

sdk-setup/src/mb2
vigejolla: Sure, if that's how you'd prefer it

sdk-setup/src/mb2
vigejolla: I agree, at least for now. Maybe some day when the expectation is that rpmlint doesn't fail we can then make it stricter again.

sdk-setup/src/mb2
vigejolla: As this topic was discussed in sw sync today, I felt that the general consensus was that indeed most users would want to have the option to disable rpmlint.

martyone: This feels like turning all this effort useless :)

Anyway I would offer an alternative suggestion then: derive the default value as I suggested above `BUILD_NORPMLINT=${SAILFISH_SDK_FRONTENT:+1}` and allow to use the existing `--no-check` option to suppress it under SDK chroot. I don't see strong need for disabling rpmlint if its failure is not fatal, so disabling all checks in the occasional cases should be OK.

vigejolla: > This feels like turning all this effort useless :)

Maybe. Time will tell.

> Anyway I would offer an alternative suggestion then: derive the default value as I suggested above `BUILD_NORPMLINT=${SAILFISH_SDK_FRONTENT:+1}` and allow to use the existing `--no-check` option to suppress it under SDK chroot. 

Now that would be clever! https://github.com/sailfishos/sailfish-qtcreator/pull/522 could then be discarded, right?

martyone: yes

martyone: The initialization is good now, just the option handling and docs is still left there.

sdk-setup/src/mb2
```diff
@@ -2672,6 +2679,12 @@ run_build() {
     local hook_exports=()
     local rpms=$(find_rpms "$OPT_OUTPUTDIR" "$OPT_PKGDIR" "$STATEDIR/spec")
     run_hook post-package "${hook_exports[*]}" ${rpms} || fatal "The 'post-package' hook failed"
+
+    if [[ ! $BUILD_NORPMLINT ]]; then
+        CHECK_SUITES="$CHECK_RPMLINT"
+        run_check__process_args_helper "" "$CHECK_SUITES" ""
+        run_check --already-locked || return 0
```
martyone, line 2686: I would rather use `return 0` as the last statement of the function, so that more steps can be added in future without changing this line.

sdk-setup/src/mb2
vigejolla: Yes, the option handling and docs were left there on purpose.

sdk-setup/src/mb2
vigejolla: fixed

sdk-setup/src/mb2
martyone: My most recent suggestion was to "... and allow to use the existing --no-check option to suppress it" and you said that would be clever :)

vigejolla: I should learn how to read

vigejolla: And yes, it is/was even more clever than I thought :)

vigejolla: better now?

sdk-setup/src/mb2
```diff
@@ -2672,6 +2674,13 @@ run_build() {
     local hook_exports=()
     local rpms=$(find_rpms "$OPT_OUTPUTDIR" "$OPT_PKGDIR" "$STATEDIR/spec")
     run_hook post-package "${hook_exports[*]}" ${rpms} || fatal "The 'post-package' hook failed"
+
+    if [[ ! $BUILD_NORPMLINT ]]; then
+        CHECK_SUITES="$CHECK_RPMLINT"
```
martyone, line 2679: one more issue. Unless I missed something, CHECK_SUITE here is not meant to refer the global variable. The global variable will be set by run_check__process_args_helper, right? In that case I would prefer having `local check_suites=...` here instead.

sdk-setup/src/mb2
martyone: great! :)

sdk-setup/src/mb2
vigejolla: :raised_eyebrow: What was I thinking? Nice catch, thanks.

martyone (approved):
I forgot to mention quotes are not needed in the assignment, so if you don't mind pushing once more... :)

vigejolla:
> I forgot to mention quotes are not needed in the assignment, so if you don't mind pushing once more... :)

removed quotes from assignment

PR #315, opened 2022-02-14
Jb57440

martyone:


sdk-setup/src/mb2
```diff
@@ -3021,7 +3021,7 @@ run_build_shell() {
             local session_start=$(date +%s)
         fi
 
-        sb2_maintain "$@"
+        sb2_maintain "$@" || return
 
         if inside_build_engine; then
             sdk-manage target sync "$OPT_TARGET" "$session_start"
```
vigejolla, line 3027: Now the sync part is not run if the maintenance command fails. Maybe it's a good thing, but I just thought I should point it out because it's a change in behavior, not mentioned in commit message.

martyone: Yup, done by intention, thanks.

vigejolla (approved):
LGTM

sdk-setup/src/mb2
martyone: But just by chance I spotted the `sdk-assistant maintain` implementation and there the sync happens even in case of non zero exit code. So we should at least synchronize them. Which one would you prefer?

vigejolla: I think running the target sync in case of failed command is almost always useless. At least when I'm imagining myself in a situation when I have accidentally made a mistake in command argument and the command fails, it would feel like extra punishment having to wait for the target sync to finish before trying again.

martyone: Updated

vigejolla (approved):

PR #317, opened 2022-02-16
[sdk-chroot] Utilize profile.d for bash setup. JB#56751

vigejolla:


vigejolla:
Please note, that this changes the order in which the files are read: Now .mersdk.profile is read as a profile file, i.e. before .bashrc. Thus setting PS1 in .mersdk.profile doesn't quite work, as .bashrc will quite likely overwrite PS1. Setting PS1 in .bash_profile, with the help of $SAILFISH_SDK, works.

martyone:
Looks promising. Doesn't it read the profile twice when opening an interactive shell?

martyone:
It might be. Some messages may get issued twice, etc., so we should really aim for single pass. Maybe just passing `-l` either to the outer shell or the inner one based on the arguments?

vigejolla:
> It might be. Some messages may get issued twice, etc., so we should really aim for single pass. Maybe just passing `-l` either to the outer shell or the inner one based on the arguments?

Definitely worth a try.

vigejolla:
Rewrote, separating the working principle depending on the arguments: When starting an interactive shell, all "magic" is done under profile.d. When executing a single command, we use the old method, with one difference though: We don't utilize a login shell. This mimics the behaviour with ssh, with one minor difference: We don't read .bashrc.

martyone:
Rebased

martyone:
Debugged, polished.

martyone (approved):

PR #318, opened 2022-02-17
Jb57498

martyone:


vigejolla (approved):
This is one of those things which at the first glance look simple. Then you start thinking about how it works in the different scenarios. And then it looks too simple to be true. But I believe you have actually tested it :)

PR #319, opened 2022-02-18
Jb57471

martyone:


sdk-setup/src/mb2
```diff
@@ -2343,8 +2343,8 @@ sign_rpms__prepare() {
 }
 
 sign_rpms() {
-    [[ -e $STATEDIR/spec ]] || fatal "No previous build found"
-    local rpms=$(find_rpms $all_opt "$OPT_OUTPUTDIR" "$OPT_PKGDIR" "$STATEDIR/spec")
+    [[ -e $TMP_SPEC ]] || fatal "No previous build found"
```
vigejolla, line 2376: Just wondering if we should pass the spec file as argument for these functions as well

sdk-setup/src/mb2
```diff
@@ -1512,6 +1512,26 @@ get_spec_tag() {
         --queryformat="$query" "$spec"
 }
 
+# Do a simple, one level expansion of %include directions. Failure to process
+# an %include directive is not considered fatal and as such is copied
+# unchanged to the output.
+flattern_spec() {
```
vigejolla, line 1518: flatten?

sdk-setup/src/mb2
martyone: yup, fixed, thanks

sdk-setup/src/mb2
martyone: Fixed

vigejolla (approved):
LGTM

PR #320, opened 2022-02-22
[sdk-chroot] Accept y/n when creating target directories. Fixes JB#57463

vigejolla:


martyone:
hm, I am just a bit worried about the option "No". It was intentionally done that the user could not proceed without having these dirs present as other tools rely on their existence.

vigejolla:
> hm, I am just a bit worried about the option "No". It was intentionally done that the user could not proceed without having these dirs present as other tools rely on their existence.

That is a silly thing to rely on, as the user can also remove the directories right after entering the chroot.

martyone:
The point is not about preventing the user to shoot them to their own knee :) The reason for doing so is that there are some assumptions about the directory layout above the directory there the chroot exists (`../../{toolings,targets}`), which is something the user may not realize initially, extracting the chroot e.g. under `~/sdk`, and we do not want to pollute random dirs on host (`/home/{toolings,targets}` in the case with `~/sdk`) without asking ...and we can't operate without them.

vigejolla:
> The point is not about preventing the user to shoot them to their own knee :) The reason for doing so is that there are some assumptions about the directory layout above the directory there the chroot exists (`../../{toolings,targets}`), which is something the user may not realize initially, extracting the chroot e.g. under `~/sdk`, and we do not want to pollute random dirs on host (`/home/{toolings,targets}` in the case with `~/sdk`) without asking ...and we can't operate without them.

Wouldn't you think that the user might realize those directories serve a purpose when sdk-chroot asks to create them?

> https://github.com/sailfishos/sdk-setup/blob/master/sdk-setup/src/sdk-chroot#L127

OTOH, we could also print that if the user selects "No" to make it more clear.

martyone:
But what would be the outcome if we let them choose No? Let them try how broken SDK doesn't work without delay and do proper installation later? The assumption whether the user realizes those directories serve a purpose or not does not change anything about the situation - the only sensible answers are yes (to continue) and abort.

vigejolla:
Removed the possibility to continue without creating directories.

martyone (commented):
```
SDK targets location '...' does not exist - about to create it.
Yes, No? [y/n] (y)
```

hmmm, how about the following?

```
SDK targets location '...' does not exist - about to create it.
Continue? [y/n] (y)
```

vigejolla:
> Continue? [y/n] (y)

I agree, that's better

martyone (approved):
There seems to be one unwanted removal of a blank line.

vigejolla:
> There seems to be one unwanted removal of a blank line.

Fixed, thanks!

PR #321, opened 2022-02-24
[sdk-manage] Use sdk-register from /usr/libexec if possible. JB#57487

vigejolla:


sdk-setup/src/sdk-manage
```diff
@@ -651,6 +651,14 @@ object_zypper() {
     return $rc
 }
 
+which_sdk_register() {
+    local object=$1
+    local sdk_register=$(enter "$object" which sdk-register)
+    enter "$object" bash -c "[[ -x /usr/libexec/sdk-register ]]" && \
+        sdk_register=/usr/libexec/sdk-register
+    echo $sdk_register
```
martyone, line 659: I was rather thinking about using /usr/libexec/sdk-setup/sdk-register as the location (not an issue that sdk-register lives in a separate package, provided that we do not use it on devices)?

Independently on that, I'd also suggest to do:

```suggestion
    enter "$object" bash -c 'PATH=/usr/libexec/sdk-setup:$PATH which sdk-register'
```

vigejolla: > I was rather thinking about using /usr/libexec/sdk-setup/sdk-register as the location (not an issue that sdk-register lives in a separate package, provided that we do not use it on devices)?

Sure, I suppose that makes sense. sdk-setup and sdk-register are quite closely related.

sdk-setup/src/sdk-manage
```diff
@@ -681,14 +686,15 @@ register_object() {
     local object= domain=
     for object in "${objects[@]}"; do
         object=$type:$object
-        domain=$(enter "$object" sdk-register -d) || return
+        sdk_register=$(which_sdk_register "$object")
```
martyone, line 689: hmm, only now I realized we could do even better by generalizing `which_sdk_register` as `object_which` and then

```suggestion
        sdk_register=$(object_which "$object" sdk-register)
```

martyone (commented):
One more suggestion and please quote `$sdk_register`

martyone (approved):

PR #323, opened 2022-03-04
Jb57663

martyone:


vigejolla (commented):
Thinking about this more. I don't really like this approach because it will result in people leaving their passwords in a plain text file. While they can remove the file after they have finished their business, I doubt many will. I would consider prompting for the username and password on the sfdk side, writing the information to a file there, which would then be used in sdk-manage, and then removed after finishing with sdk-manager. The password would still be in plain text file during the operation, but it would be removed immediately so the risk would be smaller.

Thaodan:
> Thinking about this more. I don't really like this approach because it will result in people leaving their passwords in a plain text file. While they can remove the file after they have finished their business, I doubt many will. I would consider prompting for the username and password on the sfdk side, writing the information to a file there, which would then be used in sdk-manage, and then removed after finishing with sdk-manager. The password would still be in plain text file during the operation, but it would be removed immediately so the risk would be smaller.

Since we already use gpg we could also advice people to use an encrypted netrc.

vigejolla:
> Since we already use gpg we could also advice people to use an encrypted netrc.

Does curl support that?

PR #324, opened 2022-03-04
Jb57615

martyone:


martyone:
The very first two commits belong to other PR.

sdk-setup/src/mb2
```diff
@@ -51,7 +51,32 @@ usage() {
...
+   build-requires  Update build time dependencies
+   prepare  Prepare sources using the recipe from spec file
+   apply  Apply patches listed in the spec file
+   build  Execute all build steps at once
+   qmake  Execute qmake build step
+   cmake  Execute cmake build step
+   make  Execute make build step
+   make-install  Execute make-install build step
+   package  Execute package creation build step
+   build-shell  Execute custom build steps
```
vigejolla, line 69: I think the term "custom build step" is a bit misleading here. The other build steps are defined in RPM spec files, while build-shell command can be used for executing arbitrary commands. They don't have to be build steps. You don't even need a spec file.

sdk-setup/src/mb2
```diff
@@ -51,11 +51,275 @@ usage() {
     less --quit-if-one-screen <<EOF
 $(synopsis)
 
-Executes all or a subset of build commands in the context of an rpmbuild.
+mb2 provides convenient interface to Scratchbox 2, rpmbuild and more.
 
-<specfile> will be looked for in the current rpm/ dir. If there is more than
-one it must be provided.  $ME is aware of spectacle and will update the spec
-file if there is an obvious yaml file which is newer.
+COMMANDS OVERWIEW
```
vigejolla, line 56:
```suggestion
COMMANDS OVERVIEW
```

vigejolla (changes requested):
One typo fix, also please consider if you could improve the `build-shell` overview.

sdk-setup/src/mb2
martyone: I don't see a better option. See how it aligns with Qt Creator > Projects > Build > Build Steps > Add Build Step > {qmake, Make, ..., Custom Process Step}, and see how it aligns internally - the user is allowed to execute standard build steps which operate according to .spec, and they are also allowed to executed custom build steps. Whatever is done with the help inside the build shell is considered a build step as building software is the use case we are after here, as the name also suggests.

sdk-setup/src/mb2
martyone: Fixed, thanks

vigejolla (approved):

PR #331, opened 2022-08-02
[mb2] Fix passing arguments to make with %make_build. JB#58377

martyone:


vigejolla (commented):
If I understand this correctly, the commit message describes the situation *before* the fix. Wouldn't it be less confusing to describe the fix instead?

martyone:
Updated commit message.

vigejolla (approved):
LGTM

PR #334, opened 2022-10-04
[sdk-assistant] Hide snapshots in the listing unless told otherwise. JB#58916

martyone:


sdk-setup/src/mb2
```diff
@@ -308,7 +308,7 @@ build target and using the clone to persist the redefined clean state.
 
    $ sdk-assistant clone SailfishOS-4.2.0.21-{aarch64,MyDevice}
    $ sdk-assistant maintain SailfishOS-4.2.0.21-MyDevice
-   ... add/rm packages etc. ...
+   ... add/rm repositories, packages etc. ...
```
vigejolla, line 311: I would actually leave out packages here. Adding device-specific repositories to the target makes sense, but I think in most cases adding packages should be done in snapshots, as otherwise it might be difficult to return to the clean state. Sure, sometimes it might make sense to add packages to the target, but that should be the exception and not the rule.

vigejolla (approved):
One suggestion, otherwise lgtm

sdk-setup/src/mb2
martyone: Agreed, you just missed that this particular example is exactly about your suggestion :) See that it clones the (very) original target and uses the clone to define a new clean state by preinstalling adaptation-specific packages (from adaptation-specific repos).

PR #335, opened 2022-10-18
[mb2] Hint about offending state directory in build-init help. JB#58848

vigejolla:
The state directory may have been left in a wrong place in the past, it makes sense to hint the user that it's safe to remove it.

### DIFF
--- a/sdk-setup/src/mb2
+++ b/sdk-setup/src/mb2
@@ -487,6 +487,10 @@ COMMANDS
         be initialized.  Otherwise the current working directory is treated as
         the project directory.
 
+        Initializing a build directory creates a hidden directory called
+        '.mb2'. Removing this directory stops mb2 treating the containing
+        directory as a build directory.
+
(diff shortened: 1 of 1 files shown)


sdk-setup/src/mb2
```diff
@@ -487,6 +487,12 @@ COMMANDS
         be initialized.  Otherwise the current working directory is treated as
         the project directory.
 
+        Note: Initializing a build directory creates a state directory called
+        '.mb2'. If you see error messages instructing you to run commands
+        from top of the build tree, and the mentioned directory is not your
+        build directory, it's because the state directory has been left there
+        in the past. The offending state directory can be safely removed.
```
martyone, line 494: I would prefer to shorten a bit and drop "Note:" - every line counts here :)

```suggestion
        Initializing a build directory creates a hidden directory called
        '.mb2'. Removing this directory stops mb2 treating the containing
        directory as a build directory.
```

martyone (approved):
One typo noticed, otherwise LGTM

PR #336, opened 2022-11-10
[mb2] Do not fail on multiple SPECs if SPEC is not needed. JB#59287

martyone:


vigejolla (approved):
LGTM. Although the commit message could have opened up the build-requires logic a bit. It took me a while to figure it out. Well, maybe it's just because it's morning and I haven't yet had my third cup of coffee...

martyone:
Similarly here it was not easy to write something by the end of the day :) Added a bit more info now.

PR #337, opened 2022-11-14
[mb2] Unify references to RPM SPEC files. JB#58855

vigejolla:


sdk-setup/src/mb2
```diff
@@ -1363,7 +1363,8 @@ try_to_make_spec() { # Tries to create a missing spec
     [[ -f "$1" ]] && return # It's not missing
     local yaml="${1%.spec}.yaml"
     [[ -f "$yaml" ]] || return # No yaml
-    ANSI_COLORS_DISABLED=1 specify -n -N "$yaml" || fatal "Failed to convert .yaml to .spec"
+    ANSI_COLORS_DISABLED=1 specify -n -N "$yaml" \
+        || fatal "Failed to convert YAML to SPEC file"
```
martyone, line 1367: Messages are without context, so maybe saying RPM with SPEC would be better. Applies to all messages.

sdk-setup/src/mb2
```diff
@@ -58,8 +58,8 @@ COMMANDS OVERVIEW
 
    build-init  Initialize build directory
    build-requires  Update build time dependencies
-   prepare  Prepare sources using the recipe from spec file
-   apply  Apply patches listed in the spec file
+   prepare  Prepare sources using the recipe from RPM SPEC file
+   apply  Apply patches listed in the SPEC file
```
martyone, line 62: In this kind of listing I would prefer to say RPM with every occurrence of SPEC.

martyone (approved):
Two suggestions. Please check sfdk side too - I realized some of these are still there as well.

vigejolla:
> Two suggestions. Please check sfdk side too - I realized some of these are still there as well.

Both of your suggestions were such that I already considered those already. In both cases I ended up preferring the shorter version, but I don't have strong opinions here, so I can just do as you suggested. And I'll check sfdk as well like you suggested.

vigejolla:
> Two suggestions. Please check sfdk side too - I realized some of these are still there as well.

Fixed these. Also changed the output of check --list-suites command, which I had somehow missed earlier.

martyone (approved):

PR #339, opened 2022-11-23
[mb2] Ignore %prep changes when setting version. JB#58944

martyone:


sdk-setup/src/mb2
```diff
@@ -1853,7 +1856,7 @@ maybe_fix_package_version() {
         fatal "Cannot save state of Git submodules: git-stash failed for some."
     fi
 
-    if [[ $(git_ rev-parse "$tag^{}") != "$head" || $modified_submodules ]]; then
+    if [[ $head != "$tag_sha1" || $modified_submodules ]]; then
```
vigejolla, line 1942: Why is "$tag_sha1" quoted but "$head" is not?

sdk-setup/src/mb2
```diff
@@ -1808,6 +1812,48 @@ git_commit_brief() {
...
+
+recent_patches() {
+    local prune=$([[ $1 == --prune ]] && echo 1)
+
+    if [[ $prune ]]; then
+        git_ notes --ref "$PATCHSET_ID_REF" prune
+        git_ submodule foreach --quiet --recursive "git notes --ref '$PATCHSET_ID_REF' prune"
+    fi
+
+    local reasonable_limit=20
```
vigejolla, line 1836: Why 20? To me it seems recent_patches() is only used for determining if there are any recent patches, so one should be enough?

vigejolla (approved):
LGTM.

sdk-setup/src/mb2
martyone: While word splitting and pathname expansion are not performed on the words between the `[[` and `]]`, so there is no need for quoting between these in general, a string to the right of any of the `==/!=/=~` operators is considered a pattern and matched, so (parts of) it needs to be quoted unless this is desired.

sdk-setup/src/mb2
martyone: This is used to detect whether patches have been applied, in which case it refuses to prepare again, so it needs to look behind possible commits created since that time.

PR #342, opened 2022-12-05
[mb2] Fix determining setup dir which is a sub-subdir. JB#59561 

martyone:


vigejolla (approved):
LGTM. I guess. Writing sed scripts might be fun, reading them, not so much.

PR #344, opened 2022-12-12
[mb2] Add '--stable' option to the 'scrape' command. JB#59604

martyone:


vigejolla (approved):
Functionalitywise LGTM. I would still consider changing the option to e.g. `--keep-filenames`. Something that would give a bit more clue about what is going to happen. Yes, it's mentioned in the documentation, but still I would like to have something a bit closer to what actually happens than just "stable", which could mean anything.

PR #348, opened 2023-01-05
[sdk-manage] Fix configuring http proxy for targets. Fixes JB#59759

vigejolla:


martyone (approved):
Hm, I did some final "optimizations" at the end and missed the difference of the name here and in the tooling case :\

PR #349, opened 2023-01-11
[init.container] Detect host OS reliably. JB#59670

martyone:


vigejolla (approved):
LGTM

sdk-setup/src/init.container
```diff
@@ -32,6 +32,15 @@ set -o nounset
 LOG=/var/log/systemboot.log
 NOTE="Autogenerated by $0. Do not edit!"
 
+# Backward compatibility with older sfdk - to be dropped later
+if [[ ! ${SAILFISH_SDK_HOST_OS:-} ]]; then
+    if [[ -e /dev/fuse ]]; then
```
Thaodan, line 37: /dev/fuse also exists in Linux. Does docker reveal the target plat form in a another way?

martyone: Yes, on Linux the default content of /dev is different under Docker. This changes now with the related PR under qtc where we bind mount /dev from host.

PR #351, opened 2023-02-08
[sdk-init] Separate spec files per builder. JB#60112

vigejolla:


sdk-setup/src/sdk-init
```diff
@@ -151,6 +151,14 @@ install_template()
...
             target=$OPT_NAME.pro
             ;;
+        */template.spec.qmake)
+            [[ $OPT_BUILDER == qmake ]] || return 0
+            target=rpm/$OPT_NAME.spec
+            ;;
+        */template.spec.cmake)
+            [[ $OPT_BUILDER == cmake ]] || return 0
+            target=rpm/$OPT_NAME.spec
+            ;;
```
martyone, line 161: I would consider to use names in the form of "template-<builder>.spec" so that text editors recognize the file type.

martyone (approved):
LGTM with one suggestion. Did the self-test pass after the changes?

sdk-setup/src/sdk-init
vigejolla: > LGTM with one suggestion. Did the self-test pass after the changes?

"All tests passed".

Btw. We should really catch this kind of issues on autotester; I'll look into extending the test cases next.

martyone (approved):

PR #352, opened 2023-02-13
Use /usr/local/bin/git if available. Fixes JB#60130

vigejolla:


sdk-setup/src/mb2
```diff
@@ -2688,7 +2688,10 @@ init_patch_wrapper() (
     local patchset_id=$2
 
     # Use git from host
-    git() { SBOX_DISABLE_MAPPING=1 /usr/bin/git "$@"; }
+    git() {
+        local original_git=$(which git)
```
martyone, line 2692: The `which` command needs to be used on host to be always accurate, so we need

```
HOST_GIT=$(quote "$(which git)")
```

below (preferably just above the `WE=$(quote "$WE")` line https://github.com/sailfishos/sdk-setup/pull/352/files#diff-ae2d3f30654b05a59e4f75cd131205c52ce1b713c879047de93d8965384ca516R2839) and then use `$HOST_GIT` in place of `$original_git` here.

vigejolla: That didn't quite work, as git() is used already in init_patch_wrapper, and not just in the wrapper. Also, while applying your proposed change, I noticed that init_commit_proposed_changes_hook needs the same change as well.

sdk-setup/src/mb2
```diff
@@ -2686,9 +2686,10 @@ END
 init_patch_wrapper() (
     local wrapper=$1
     local patchset_id=$2
-
+    HOST_GIT=$(which git)
```
martyone, line 2689:
```suggestion
    local host_hit=$(which git)
```

sdk-setup/src/mb2
```diff
@@ -2941,9 +2943,10 @@ switch_to_snapshot() {
 init_commit_prep_changes_hook() {
     local hook=$1
     local patchset_id=$2
+    HOST_GIT=$(which git)
```
martyone, line 2946:
```suggestion
    local host_hit=$(which git)
```

martyone (commented):
Yeah one can get easily lost in the various execution contexts :) Alternatively we could use the `command git` in the initialization code, but both options work for me.

sdk-setup/src/mb2
vigejolla: now if we use `local host_git` here (I assume "hit" was a typo), we need to change git() to use `"$host_git"?` (lower case), which means we need to use lower case also in the wrapper. Are you ok with that?

martyone: hm yeah, the use in the initialization again... so I would return to my original suggestion and use `command git` in the initialization code to avoid the issue you hit, i.e., to make it use the real git instead of the wrapper function in that context.

martyone (approved):

PR #356, opened 2023-03-03
[mb2] Ensure that mb2 apply will work with tabs in spec file. JB#60285

Thaodan:


vigejolla (approved):
I'm not sure how good idea it is to suddenly start using tabs instead of spaces in spec files. But since it's not against the specification of the spec files, it's probably better to support them in mb2.

PR #359, opened 2023-12-18
[mb2] Test also tabs in spec file in self-test. JB#60285

vigejolla:


sdk-setup/src/mb2
```diff
@@ -6492,7 +6492,7 @@ License:    LICENSE
 URL:        http://example.org/
 Source0:    %{name}-%{version}.tar.bz2
 # rpmbuild applies patches in the order they appear here, ignoring their numbering
-${enable_patch:+Patch1:  test-fix-1.patch}
+${enable_patch:+Patch1:    test-fix-1.patch}
```
pvuorela, line 6495: If this is an intentional hard tab, could have some note that it's supposed to be there. Those are quite easy to turn into normal spaces by accident.

pvuorela (approved):

PR #361, opened 2024-06-05
Fix generating QML bundle for multiline entries. JB#59991

martyone:


martyone:
Just the second commit belongs to this PR

martyone:
Wrong bug ID, replaced with PR #362 .

PR #362, opened 2024-06-05
Fix generating QML bundle for multiline entries. JB#62194

martyone:


martyone:
Just the second commit belongs to this PR

keto (approved):

PR #363, opened 2025-10-05
Handle %patch N and %patch -P N macro usage too

mlehtima:


sdk-setup/src/mb2
```diff
@@ -3253,7 +3253,8 @@ run_apply() {
     local to_apply=
     local auto_applied=$(grep -q "^%autosetup" "$TMP_SPEC" \
         && ! grep -q "^%autopatch" "$TMP_SPEC" \
-        && ! grep -q "^%patch[0-9]" "$TMP_SPEC" \
+        && ! grep -q "^%patch\s?[0-9]" "$TMP_SPEC" \
+        && ! grep -q "^%patch.*-P [0-9]" "$TMP_SPEC" \
```
keto, line 3257: Could there be more than one space after -P? or no space at all?

sdk-setup/src/mb2
keto, line 3263: Need to modify this too?

PR #366, opened 2026-04-24
[mb2] Allow to customize dirty version suffix. JB#59271

martyone:


sdk-setup/src/mb2
```diff
@@ -983,7 +983,9 @@ GLOBAL OPTIONS
         denoted by the (latest or selected) tag, a suffix composed from the
         current branch name, time stamp and a commit SHA1 will be added to the
         package version. If git-state is not clean a git-stash will be created
-        and its SHA1 will be used.
+        and its SHA1 will be used. Similarly, an SHA1 will be recorded for each
```
pvuorela: I'd spell out SHA1 as "sha one", thus 'a' would feel more appropriate than 'an'.

pvuorela (approved):

sdk-setup/src/mb2
martyone: Resorted to say "hash" instead, as SHA-1 is not the only option anymore.

